PR #3, opened 2020-12-25
Docs: More work.

bjones1:


.flake8
```diff
@@ -16,6 +18,7 @@ ignore=F821,
   # web2py needs to compare to ``== True`` or ``== False`` for queries.
   E712
 
+# :index:`question`: **Why are these here?** I thought that moving away from web2py means we can get rid of this horrible stuff...
```
bnmnetp, line 21: because I just copied the file.

app/routers/books.py
```diff
@@ -20,32 +25,40 @@
...
-# Note the use of the path type for filepath in the decoration.  If you don't use path it
-# seems to only get you the ``next`` part of the path /pre/vious/next/the/rest
+# any static file contact with web2py and handle static files upstream with nginx directly; therefore, this is useful only for testing/a non-production environment.
+# Note the use of the ``path``` type for filepath in the decoration.  If you don't use path it
+# seems to only get you the ``next`` part of the path ``/pre/vious/next/the/rest``.
+#
+# :index:`todo`: **Routes for draft (instructor-only) books.**
 @router.get("/published/{course:str}/_static/{filepath:path}")
 async def get_static(course: str, filepath: str):
+    # :index:`fixme`: **Find a way to avoid a hard-coded path.** This fix is needed for the route below as well.
```
bnmnetp, line 59: This is still a prototype.... I think should make a configuration variable like BOOK_LIBRARY  that points to a folder that holds all of the books.  The great thing is it does not have to be inside the RunestoneServer folder!

app/routers/books.py
```diff
@@ -20,32 +25,40 @@
...
+# Routing
+# =======
 # Setup the router object for the endpoints defined in this file.  These will
-# be connected to the main application in main.py
-#
+# be `connected <included routing>` to the main application in `../main.py`.
 router = APIRouter(
-    prefix="/books",  # shortcut so we don't have to repeat this part
-    tags=["books"],  # groups all logger tags together in the docs
+    # shortcut so we don't have to repeat this part
```
bnmnetp, line 38: Is there any other option for including these comments?  I find this style to be very distracting.  To me it makes the function call a lot harder to read and understand.  Also, stylistically we should decide, do we need to repeat this in every router file?  I notice that you left rslogger.py alone.  I know someone could dive in to any given router file, so it doesn't make sense to just do it in one file.  But too much repetition is not good either.

pyproject.toml
```diff
@@ -14,6 +14,7 @@ fastapi = "^0.62.0"
 uvicorn = "^0.13.1"
 Jinja2 = "^2.11.2"
 aiofiles = "^0.6.0"
+# :index:`fixme`: shouldn't we install a `backend <https://www.encode.io/databases/#installation>`_?
```
bnmnetp, line 17: So far I've been working with sqllite.  But we should also install what we need for postgresql when we get to that.

bnmnetp, line 17: It would be nice if we could do :index:`fixme:` or :index:`FIXME:` so that these tags are compatible with the usual editor plugins.

app/routers/books.py
```diff
@@ -20,32 +25,40 @@
...
+# any static file contact with web2py and handle static files upstream with nginx directly; therefore, this is useful only for testing/a non-production environment.
+# Note the use of the ``path``` type for filepath in the decoration.  If you don't use path it
+# seems to only get you the ``next`` part of the path ``/pre/vious/next/the/rest``.
+#
+# :index:`todo`: **Routes for draft (instructor-only) books.**
 @router.get("/published/{course:str}/_static/{filepath:path}")
 async def get_static(course: str, filepath: str):
+    # :index:`fixme`: **Find a way to avoid a hard-coded path.** This fix is needed for the route below as well.
+    #
+    # :index:`todo`: **Are there any security concerns?** Could a malicious user pass a "path" of ``../../../private_files`` and access anything on the server?
```
bnmnetp, line 61: The browser subtracts the ../ from the URL path before sending it to the server. So we get a 404 and the get_static endpoint is not even called.

We could -- just to be safe -- make sure to either reject a relative path or make sure that our file path resolves to something below whatever we define as our BOOK_LIBRARY

index.rst
```diff
@@ -3,16 +3,26 @@ BookServer documentation
...
 
+
+Documentation
+=============
 .. toctree::
     :maxdepth: 1
 
     docs/installation
     docs/goals
+    docs/design
```
bnmnetp, line 16: the descriptions of the files I have in here may be better to add to the toctree in the app folder??  That would keep it from being disjointed.

I would like this to be readable and useful.  So maybe spending some time up front thinking about where we put detail would be useful.

.flake8
bjones1: :) I'll take them out. My goal is to do a code review as I add docs, so here's progress!

Merry (late) Christmas!

app/routers/books.py
bjones1: That makes sense. I'll leave the comment in so we can find this when it's time to fix. Just more code review comments...

app/routers/books.py
bjones1: Good questions. The CodeChat system can't support mixed comments and code -- [here's](https://codechat.readthedocs.io/en/master/CodeChat/CodeToRest.py.html#mixed-code-and-comments) a (somewhat obscure) explanation why. So, when using CodeChat, avoid mixed comments and code.

Given that, I'm open to other styles -- [here](https://codechat.readthedocs.io/en/master/docs/style_guide.cpp.html#an-unindented-heading-correct) is my style guide and reasoning for the style I use. What would make more sense to you?

app/routers/books.py
bjones1: Sounds good. I try to be a bit paranoid about potential security issues. I could also argue that there's no need to fix this, since a production server should never be serving files this way. But...I think a relative path check or something like that would be best. I'd suggest the equivalent to [safe_join](https://flask.palletsprojects.com/en/1.1.x/api/#flask.safe_join) -- there's a copy of the relevant code in web2py's `books.py` starting at line 344.

index.rst
bjones1: I went back and forth on this -- at first, the were part of the toctree. Then, at three files, it seems better to have them under a separate heading. However, if we combine goals and design, I can see putting these back together.

I'll put them back in the same heading as the code for now, and we can always move them based on what seems to provide a good flow.

pyproject.toml
bjones1: On VSCode, making that `` :index:`FIXME` `` works. How about that?

bjones1: Should I therefore make this `pip install databases[sqlite]` for now?

app/routers/books.py
bjones1: To repeat something, we can use reST replacement syntax. Another (perhaps better) option: document it in one place, then hyperlink to that from other similar code.

I'll do that, then tell me what you think.

app/routers/books.py
bjones1: Do you want me to add this?

bjones1:
The previous push provides updates based on your feedback.

pyproject.toml
bnmnetp: If databases[sqlite,postgresql] works that is maybe the best option?  As of now I have explicitly installed what we need for both, but perhaps leaving those as somewhat hidden dependencies is better than calling them out.

app/routers/books.py
bnmnetp: Sure.

While you are in code review mode...  One question I've been thinking about is whether we need the `db` parameter on the function in crud.py or whether we should just import `db` into crud and use it.?

Also BTW, everything defined as a setting in config.py can be set by an environment variable.  Where the environment variable is the upper case version of the string in the settings class.  If you want to add book_library or book_folder as a setting to act as the base directory for books that would be great.

bnmnetp: Also, style-wise - sqlalchemy can use the bitwise operators just like DAL for and and or OR we can use the functional equivalent `and_` and `or_` in sqlalchemy, but I think we should pick one and be consistent.

pyproject.toml
bjones1: It looks like `databases = {extras = ["sqlite", "postgresql"], version = "^0.4.1"}` works. I'll add that and remove the `aiosqlite` entry.

app/routers/books.py
bjones1: I think importing `db` is fine.

I'll think about nice ways to specify a path.

I strongly prefer operators like DAL -- the `and_` and `or_` functions are quite awkward.

bjones1: Done re: safe_join and book_path.

PR #4, opened 2020-12-28
Fixes per PR discussion (safe_join, book_path)

bjones1:


bjones1:
I've completed a full first pass through the code and left lots of comments/ideas. What do you think about this approach (placing questions in an `:index:` to make them easy to walk through? I like the github code review, but find this more convenient to provide feedback as I read through the code in my local editor.

Warning/danger: the tests are failing for me. Is this my setup, or do they fail for you?

Do you mind if I delete unused imports?

Have you tried to run the `pre_commit_check,py` script? I like it as a good way to make sure the code is in good shape before a commit.

bjones1:
Also, have you considered using rebases instead of merges? I like that -- it's much cleaner and easier to follow in git.

bnmnetp:
Thanks,

The unit tests are failing for me, for multiple reasons.

1.  We need to ignore _build, otherwise when pytest collects questions it finds a duplicate test_rslogging.py 
2. I updated the schema for incoming data but not the tests.
3. After fixing 1 and 2 the test we have triggers the scenario with ans_idx that you point out.

At this early stage I wasn't too worried as a lot of this is experimental / early stage just trying to see how things fit together.

I don't mind you deleting the unused imports but I was not doing so myself despite their annoying appearance from flake8 in my editor because I was anticipating that I would need to add them right back in again in most cases.

I have not tried the `pre_commit_check.py` script.    At some point (very soon) I will stop treating this as an experiment/learning laboratory and get down to business. 😄  Including not committing directly to master, running the tests regularly setting up a workflow to run the tests. etc.

To be honest I'm not a fan of the idea of using `:index:` for the walkthrough as I can't respond without modifying the code and committing it.  The code review features of GitHub are great because we can have a discussion about things without muddying the code and the history with lots of back and forth.

bnmnetp:
Regarding rebasing -- I don't use rebase very often.  I think it was a learned behavior from years ago when nearly every time I tried to rebase I think I did it wrong or it didn't work how I thought it should.  Whereas merge always did exactly what I expected even if it made the history more cluttered.

This article https://www.atlassian.com/git/tutorials/merging-vs-rebasing#the-golden-rule-of-rebasing  Especially the Golden Rule makes me hesitant to adopt this strategy.  I confess it always makes me nervous when I see you force pushing things. Which I think is exactly what the Golden Rule section of that article is talking about.

I often find myself pulling a PR to a local branch and testing it before approving.  During testing I sometimes make a minor tweak before merging everything back to master and then pushing to accept the PR.  This is where I think I can get into trouble with the rebasing and force pushing.  I've noticed that once I have the branch I cannot just do another pull from that branch and get updates if you have force pushed.  I haven't done a thorough investigation of this but that seems to be my experience.  I have to delete the branch and pull it fresh.

So... In thinking about guidelines and workflow for this new repository. What do you think of this?
0. No committing to main directly.  I will do my work on branches and make PRs
1. In principle I like the idea of rebasing a branch I'm working on locally.
2. But after I publish that branch and make it a PR I would say no more rebasing after the branch is public?
3. The main branch should be protected on GitHub so that tests must pass and have an approving review from at least one committer.  Ideally this would be someone other than the person who makes the PR but in practice this is hard when you get busy and don't have time to review.  Plus I think its fine for self review in small simple cases or emergency bug fixes.
4. Using your script or setting up some pre commit hooks is a great idea and a great way to help others follow our guidelines.
5. Tests should be required part of a PR -- we have the opportunity to get our coverage at 100% in these early stages and keep it there as we build out this new server.
6.

app/crud.py
```diff
@@ -70,6 +73,7 @@ async def create_answer_table_entry(db: Database, log_entry: schemas.LogItem):
     return res
 
 
+# :index:`TODO`: **I think the idea here**, but the implementation will still need some special cases for getting the specific data for all the question types.
```
bnmnetp, line 76: My thinking here is that we always return the entire row for any given question so that it **does** have any extra columns added for a particular question type.

app/crud.py
```diff
@@ -59,9 +60,11 @@ async def create_answer_table_entry(db: Database, log_entry: schemas.LogItem):
     values = {
         k: v
         for k, v in log_entry.dict().items()
+        # :index:`question`: **Why exclude** ``act``? Some types (clickablearea, shortanswer, etc) use this field. There's probably more conditional logic needed here based on the event type.
```
bnmnetp, line 63: `act` is not stored in any of the answer tables, as `act` has always been a sort of catch-all field for the use info table that needs to be parsed in a unique way for each component.

ShortAnswer and Clickable should be updated to populate the answer field when they call logBookEvent even if it is just a duplicate of the act field.

alembic/toctree.rst
```diff
@@ -1,7 +1,9 @@
 ********************************
 Database migration using Alembic
 ********************************
-:index:`docs to write`: overall approach/use. This is a generic, single-database configuration.
+:index:`docs to write`: **overall approach/use.** This is a generic, single-database configuration.
+
+:index:`TODO`: **What's the plan for migrations?** web2py also does migrations, so how do we avoid a migration duel between web2py and Alembic?
```
bnmnetp, line 6: Great question.  What I love about alembic is that migrations are intentional, both in their creation and in their applicaition.  Whereas web2py wants to do migrations automatically. ( I have them disabled for production and development and use `rsmanage migrate` myself)

We have definitely added some extra work for ourselves here, but I don't know if we can avoid it as we are building for two important use cases.

1. A new book server for our large live public systems (runestone.academy)
2. An easy to use replacement for the current `runestone serve`

If all we cared about was  1, we could just have sqlalchemy slurp up the table definitions from the database.  But that does not work for 2.  

I think the way we need to proceed is to be very careful about schema changes and when they are necessary we make sure we do them in both places.

app/schemas.py
```diff
@@ -23,21 +21,21 @@
...
+# Schemas
+# =======
 class LogItemIncoming(BaseModel):
     """
     This class defines the schema for what we can expect to get from a logging event.
     Because we are using pydantic type verification happens automatically, if we want
     to add additional constraints we can do so.
     """
 
+    # :index:`TODO`: Is there any way to specify max string lengths? The database has fixed-length fields for some of these.
```
bnmnetp, line 38: Schemas contains Pydantic models.  There is a plugin for FastAPI to allow one to generate Pydantic schemas from Sqlalchemy models, I think we ought to investigate using that.  Otherwise there is some inevitable duplication of work.  But I do really like using these schemas and probably have a lot more to learn about them.

bjones1:
> The unit tests are failing for me, for multiple reasons.
> 
> 1. We need to ignore _build, otherwise when pytest collects questions it finds a duplicate test_rslogging.py

My `pre_commit_check.py` script runs tests in the `test/` subdirectory, and I run them by hand the same way -- this is probably the easiest way to "ignore" `_build/`.

> 2. I updated the schema for incoming data but not the tests.
> 3. After fixing 1 and 2 the test we have triggers the scenario with ans_idx that you point out.
> 
> At this early stage I wasn't too worried as a lot of this is experimental / early stage just trying to see how things fit together.

That makes sense; I wasn't sure if my changes had broken something, or it was just early stage code.

> 
> I don't mind you deleting the unused imports but I was not doing so myself despite their annoying appearance from flake8 in my editor because I was anticipating that I would need to add them right back in again in most cases.

Sounds good. I'll leave them alone.

> 
> I have not tried the `pre_commit_check.py` script. At some point (very soon) I will stop treating this as an experiment/learning laboratory and get down to business. 😄 Including not committing directly to master, running the tests regularly setting up a workflow to run the tests. etc.

Great!

> 
> To be honest I'm not a fan of the idea of using `:index:` for the walkthrough as I can't respond without modifying the code and committing it. The code review features of GitHub are great because we can have a discussion about things without muddying the code and the history with lots of back and forth.

Sounds good. I'll switch to performing code review here.

> Regarding rebasing -- I don't use rebase very often. I think it was a learned behavior from years ago when nearly every time I tried to rebase I think I did it wrong or it didn't work how I thought it should. Whereas merge always did exactly what I expected even if it made the history more cluttered.

There's a balance in force-pushing between a clean history and confusing others. As you say, drawing the line at a public push is a good place to find the balance. Rebases are much easier to follow in history that merges.

> This article https://www.atlassian.com/git/tutorials/merging-vs-rebasing#the-golden-rule-of-rebasing Especially the Golden Rule makes me hesitant to adopt this strategy. I confess it always makes me nervous when I see you force pushing things. Which I think is exactly what the Golden Rule section of that article is talking about.
>
> I often find myself pulling a PR to a local branch and testing it before approving. During testing I sometimes make a minor tweak before merging everything back to master and then pushing to accept the PR. This is where I think I can get into trouble with the rebasing and force pushing. I've noticed that once I have the branch I cannot just do another pull from that branch and get updates if you have force pushed. I haven't done a thorough investigation of this but that seems to be my experience. I have to delete the branch and pull it fresh.

Sigh. I see what you mean. It's so nice to force-push on a PR to clean up incorrect commits, but it's painful for others doing a review. I'll try to be better-behaved in that respect.

> So... In thinking about guidelines and workflow for this new repository. What do you think of this?
> 0. No committing to main directly. I will do my work on branches and make PRs
> 
> 1. In principle I like the idea of rebasing a branch I'm working on locally.
> 2. But after I publish that branch and make it a PR I would say no more rebasing after the branch is public?
> 3. The main branch should be protected on GitHub so that tests must pass and have an approving review from at least one committer.  Ideally this would be someone other than the person who makes the PR but in practice this is hard when you get busy and don't have time to review.  Plus I think its fine for self review in small simple cases or emergency bug fixes.
> 4. Using your script or setting up some pre commit hooks is a great idea and a great way to help others follow our guidelines.
> 5. Tests should be required part of a PR -- we have the opportunity to get our coverage at 100% in these early stages and keep it there as we build out this new server.
> 6. As this is designed to be released as a package on pypi we ought to think about criteria for when to do a release.  I like using semantic versioning, and I've found that doing a weekly release for runestone components works well.

I agree to all this. Have you see the `dev_guide.rst`? I had some additional workflow items there. Would you update it with the items above?

alembic/toctree.rst
bjones1: I definitely prefer the idea of a deliberate, manually-invoked migration over web2py's "just do it" approach. Some thoughts and questions:

- I agree with the goal of replacing `runestone serve` in the long run, but I think we should first get the absolute minimum of endpoints working then deploy that before moving forward. IMHO, what you're currently started (`hsblog`, `getAssessResults`, and book serving) should represent a vast majority of the site traffic. I propose not adding any more endpoints until this is in production.
- This means clearly diving the database migration into "managed by Alembic" (all tables SQLAlchemy knows about) and "managed by web2py" (everything else). Then, after struggling through a migration, we need to write up a guide on how to do this correctly. My guess:
    - Update model definitions in both web2py and BookServer.
    - Shut down web2py.
    - Migrate via Alembic.
    - Re-enable web2py

app/crud.py
bjones1: OK. We'll have to look at those cases carefully -- this also means changes to `ajax.py` until we start using the BookServer.

For the future, it would be good to better document/define what each component assigns to each query parameters in the `hsblog` endpoint -- there's currently some odd behavior you mentioned going on. (`unittest` and `timedexam` also use `act`.)

app/crud.py
bjones1: That seems reasonable. I think most of this will be resolved with good tests.

app/schemas.py
bjones1: I agree. When I first started on the NewServer (Flask-based) several years ago, I spent a lot of time writing validation code, something sorely lacking in web2py. Some helpful ideas from there:

- Define a standard `id` column type and re-use it in all tables.
- I've created a Boolean data type that works with web2py/DAL's weird boolean.
- Standard columns for all answer tables we define once and re-use.
- Some [docs](https://github.com/bjones1/NewServer/blob/master/runestone/api/endpoints.py#L125) on the `hsblog` endpoint. Does this look right? If so, we should incorporate it in the current code.

app/crud.py
bnmnetp: Yep, act has always been a freeform which can be parsed based on the component.  Whereas logBookEvent just takes a javascript object that can have any fields we want populated. Yes, its a little redundant...

app/schemas.py
bnmnetp: Good ideas except for having a standard numeric id.  I honestly cringe every time I have to deal with that.  WE HAVE good logical primary keys for many of our tables and it hurts to have to use an arbitrary number rather than something meaningful just because of web2py.  Any good class on data modeling (which I have taught many times) will have you always prefer good natural primary keys over an arbitrary number.

bnmnetp: Can you integrate your Boolean data type for use with our models?  That seems like a wonderful solution.

bnmnetp:
I will update dev_guide.rst with the above.

app/schemas.py
bjones1: I'm happy to follow your lead on avoiding an ID. All my knowledge of databases is self-taught; I've never had any formal instruction. However, I think we need to select a more consistent naming convention -- for example, `course_id` is a string referring to the course name in the `useinfo` table, while it's an int referring to the id in the `code` table. This sort of nomenclature confusion makes it hard to write understandable code.

Perhaps we can plan a refactor at some point to address both of these -- ferret out numeric IDs and also move to consistent naming.

I'll drop in my bool.

bjones1: Looking at the code, I still think we need a standard numeric ID generator, not because it's a good idea for the database, but because a bunch of our tables already have this and it's good DRY practice. It will also help mark all the places that need to ID removed when the time comes to clean this up.

bnmnetp: Yes, because we need to remain backward compatible with web2py / DAL for the time being we will need to keep the numeric ids.

bnmnetp: the long term goal should be to clean up the schema as well!

PR #6, opened 2020-12-29
Organization update

bnmnetp:
* everything under bookserver directory will be in the package published to pypi
* Need a less generic name than app for the app directory
* config should be part of the package


bnmnetp:
@bjones1 -- This PR just does some reorganization so that we can use Poetry for building and publishing this book server package to PyPI.  

It now runs the pre_commit_check.py as part of CI

Oddly, and confusingly, I rebased master onto this branch after I accepted your last PR.  But now it makes it look like all of those changes are part of this PR.  That seems more confusing to me.  I thought that by doing a rebase it kept the history cleaner.  This seems more confusing as your changes are definitely not part of this PR they were part of a different PR.

bjones1:
What I see in the history isn't a rebase but an odd merge. After that merge, the rebase makes no sense -- the merge needs to be backed out, then the rebase done on top of main.

I can walk you through it if you'd like.

This PR makes sense and is what I expected -- another level of subdirectories needed for pypi...

bnmnetp:
So when you want to incorporate changes from main into your branch that you are preparing to make a PR with do you not simply do `git rebase main` ??

I don't see why I would need to back out a merge?  If this branch merges to main all is good.??

bnmnetp:
Actually I just merged on my machine and the log looks fine.  It is only on GitHub for this PR that things look so funny.

PR #8, opened 2021-04-24
WIP: review the overall structure

bjones1:
I created this PR as a way to providing a glimpse at my progress in reviewing the server. I plan on squashing a lot of these commits before it becomes a final PR.


.github/workflows/python-package.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         poetry install
     - name: flake8 quick test
       run: |
-        # stop the build if there are Python syntax errors or undefined names
+        # stop the build if there are Python syntax errors or undefined names. TODO: this is being run in ``pre_commit_check.py`` with difference parameters. Why?
```
bnmnetp, line 37: This was part of a suggested workflow that I copied.  The idea is to run a very quick test looking for syntax errors up front as their is no point in going any further if the code is broken.

bookserver/crud.py
```diff
@@ -47,6 +47,7 @@
 # -------
 async def create_useinfo_entry(log_entry: schemas.LogItemIncoming):
     new_log = dict(
+        # FIXME: not right...
```
bnmnetp, line 50: fixed but not pushed until just now.

bookserver/db.py
```diff
@@ -44,6 +44,7 @@
 else:
     connect_args = {}
 
+# TODO: databases doesn't yet support sqlalchemy 1.4, which has migration options to `sqlalchemy 2.0 mode <https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-four-use-the-future-flag-on-engine>`_. This is something we want to do from the beginning in our code!
```
bnmnetp, line 47: I know that the later version of 1.4 directly support async and await.  I'm unclear about the direction on fast api but it seems like using sqlalchemy directly with out databases in the middle would be optimal.

bookserver/schemas.py
```diff
@@ -35,7 +35,9 @@ class LogItemIncoming(BaseModel):
     to add additional constraints we can do so.
     """
 
-    # :index:`TODO`: Is there any way to specify max string lengths? The database has fixed-length fields for some of these.
+    # FIXME: Use max lengths for strings based on the actual lengths from the database using `Pydantic constraints <https://pydantic-docs.helpmanual.io/usage/types/#constrained-types>`_. Is there any way to query the database for these, instead of manually keeping them in sync?
+    #
+    # Even better, we should introspect the database and auto-convert this to Pydantic. See the idea at https://github.com/tiangolo/pydantic-sqlalchemy/blob/master/pydantic_sqlalchemy/main.py. See also https://github.com/kolypto/py-sa2schema. Other ideas: https://docs.sqlalchemy.org/en/14/core/reflection.html,
```
bnmnetp, line 40: I had looked at that too.  I don't like the idea of redoing model work if we don't have to.  A few months ago the comments were that the pydantic-sqlalchemy was not ready for prime time.  But agree that we should nail down a better solution now rather than later.

bookserver/schemas.py
```diff
@@ -61,6 +63,7 @@ class LogItem(LogItemIncoming):
     class Config:
         orm_mode = True
         # this tells pydantic to try read anything with attributes we give it as a model
+        # TODO: This doesn't make sense to me -- we're not using an ORM. What does this do?
```
bnmnetp, line 66: I don't recall.  It may be cruft some something I was experimenting with.  I don't see anywhere in  our code where we are using this Class.

conf.py
```diff
@@ -186,10 +187,12 @@
     # The ``CodeToRestSphinx`` extension creates a file named
     # ``sphinx-enki-info.txt``, which should be ignored by Sphinx.
     "sphinx-enki-info.txt",
+    # TODO: Notes here. I assume this is produced by the coverage run in the test suite? A link to that point in the code would be nice.
     "test/htmlcov",
-    "alembic/versions",
-    "**/.mypy_cache",
+    # TODO: Why ignore this?
     "bookserver/templates",
+    # TODO: notes here on how Alembic works. I think it stores migration data here.
```
bnmnetp, line 194: I think this is a weird place to document alembic.  We should have an rst file in the alembic folder that describes it and how we use it.

conf.py
```diff
@@ -186,10 +187,12 @@
     # The ``CodeToRestSphinx`` extension creates a file named
     # ``sphinx-enki-info.txt``, which should be ignored by Sphinx.
     "sphinx-enki-info.txt",
+    # TODO: Notes here. I assume this is produced by the coverage run in the test suite? A link to that point in the code would be nice.
     "test/htmlcov",
-    "alembic/versions",
-    "**/.mypy_cache",
+    # TODO: Why ignore this?
```
bnmnetp, line 192: I think there is some philosophy that we need to talk through about what we include in the document and how we organize it so that a newcomer can easily find what they need to get started and be useful.  Right now the documents we produce are a bit like looking for a needle in a haystack for important stuff.  IMHO we should not bury big picture stuff in the details of an implementation.  And we shouldn't overwhelm the user by including too many things.

bookserver/models.py
```diff
@@ -12,6 +12,8 @@
 #
 # :index:`docs to write`: It is also possible...
 #
+# TODO: What is the database strategy? Who owns/defines/creates which tables? The web2py admin app currently creates and migrates everything. But we're duplicating some tables here. Shouldn't these simply be reflected from the existing database structure, so there's only one "owner" of the table structure (web2py)? But if so, how to correctly interpret web2py booleans?
```
bnmnetp, line 15: I would prefer to have sqlalchemy own the schema.  Especially long term, and even more so because we can create appropriate indices as part of the sqlalchemy table definition.  Which we cannot do with DAL. 

I like your solution for web2py booleans in sqlalchemy, and sqlalchemy has an async/await strategy in place.

**Long term** the admin interface will be ported as well.

bjones1: Having sqlalchemy own the schema long-term makes sense. However, I don't know what a coherent "database migration" strategy looks like. [Migration meaning from web2py to FastAPI/sqlalchemy]. Especially concerning is the thought of Alembic doing migrations on some of the tables while web2py does migrations on other tables. How does this work?

I know sqlalchemy is comfortable with creating Table objects directly from a database (reflection). This makes it easy to use when web2py owns the entire schema. I don't know how/if web2py/DAL can use reflection to infer a table. Assuming it can't, we need to mark some tables as "don't migrate" to web2py -- how is this done? Assuming that's possible, I can see doing this to transfer schema ownership of some tables to sqlalchemy. We'd also need a way to tell Alembic to migrate some tables (the ones owned by sqlalchemy) but not others (owned by web2py).

All if this would be great to capture in the docs, because it will take a while to port everything.

bookserver/models.py
```diff
@@ -143,6 +145,7 @@ def copy(self, **kw):
 
 # The parsons_answers table is the only outlier in that it adds a source column to keep
 # track of which blocks were not used in the answer.
+# TODO: What about timed exams?
```
bnmnetp, line 148: I think the timed exam table is a different beast than the xxx_answers tables.

bjones1: Sounds good. I'll remove this notes.

bookserver/models.py
bnmnetp: I think this is relatively easy?!?  With DAL we can just set a table to NOT do any migrations (migrate=False). which means that DAL will continue to use its own definition of the table without even trying to reflect/and/or update the table in the database.

I *think*?? that for most things this is the right strategy.  But I agree that is a bit confusing until everything is moved over to FastAPI.

bjones1:
I'd like to sketch out an overall documentation approach with you -- I'll create a page on the Wiki for now as a jointly-edited space. Let's plan how the docs should look there, then move that structure to the code once we agree on it.

bjones1:
I wrote a first pass at the outline at https://github.com/bnmnetp/BookServer/wiki/Overall-documentation-structure. Would you review and edit this?

bookserver/models.py
bjones1: Thanks, that makes sense! Assuming that Alembic will only migrate the tables defined by this app, then this sounds good. Let's just make sure to capture this approach in the docs.

conf.py
bjones1: I started a Wiki page to help us organize big-picture docs at https://github.com/bnmnetp/BookServer/wiki/Overall-documentation-structure.

bookserver/db.py
bjones1: I posted an [issue](https://github.com/bnmnetp/BookServer/issues/9) for this.

bookserver/schemas.py
bjones1: I posted an issue for this.

.github/workflows/python-package.yml
bjones1: I went ahead and removed this check and updated flake8 checks.

conf.py
bjones1: I've updated the docs.

bookserver/config.py
```diff
@@ -45,16 +45,16 @@ class Settings(BaseSettings):
...
     test_dburl: str = "sqlite:///./runestone_test.db"
 
-    # Configure ads.
+    # Configure ads. TODO: Link to the place in the Runestone Components where this is used.
     adsenseid: str = ""
     num_banners: int = 0
     serve_ad: bool = False
 
-    # :index:`docs to write`: **What's this?**
+    # TODO: where is this used? I don't see it in the code.
```
bnmnetp, line 53: library path was an early name for book path. It can be removed

bookserver/schemas.py
bnmnetp: A little more research reminded me that this is how you configure Pydantic.    See:  https://pydantic-docs.helpmanual.io/usage/models/#orm-mode-aka-arbitrary-class-instances I'm not sure we want to remove it.  I think a little better understanding of how we are using Pydantic will make the decision.

bookserver/config.py
bjones1: OK, done in my branch.

bookserver/schemas.py
bjones1: I saw that page, but don't see how it's relevant to what we're doing. My understanding of the approach:

1. Given a sqlalchemy table, auto-generate a Pydantic schema for it.
2. From an endpoint, put data from Ajax into the validator and check it before writing to the db. The validator will also handle conversion from string to other data types if this came as query params.

bnmnetp: So, we are not doing 1 yet.  But I think we agree on that as the direction.

For 2, I think the other point is that the Schema can define what we get from the client as well as validate.  Right now in web2py you really have no way of knowing what data might be in `request.vars`. Using one of these schema objects tells you immediately what you get.  But I think that means that we will end up creating additional schema objects that do not have a one to one mapping with the database.

bjones1: I agree -- creating additional schemas for things that don't map directly to a db table makes sense. The lack of any kind of validation in web2py code produced a lot of bugs/problems.

I mainly wanted to avoid duplicating schemas when the sqlalchemy Table structure already has that data (keep the code DRY and avoid mistakes where one place is updated but the other isn't.)

bookserver/session.py
```diff
@@ -30,6 +30,7 @@ async def load_user(user_id: str) -> schemas.User:
     rslogger.debug(f"Going to fetch {user_id}")
     user = await fetch_user(user_id)
     if user:
+        # TODO: I don't understand -- why do this here? Are we validating an existing user object?
```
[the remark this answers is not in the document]
bjones1: I don't see what benefit we get from separation; the db data has already been validated / put in a known format, so that seems like what we need with additional steps.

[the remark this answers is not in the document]
bnmnetp: One benefit we do get is that a Pydantic object is mypy-compatible as well as supporting autocomplete in some editors.  Database results remain opaque.  But that may not be enough of a benefit.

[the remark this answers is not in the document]
bjones1: I need to read up more on how FastAPI integrates with sqlalchemy. I'd also like to see an example of validating HTTP parameters, then storing that in the db, to see how the code ends up looking. That will help shape the "right" way to do things.

bjones1:
I updated https://github.com/bnmnetp/BookServer/wiki/Overall-documentation-structure to add user docs and developer docs sections. Feedback would be great!

bookserver/session.py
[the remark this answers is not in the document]
bnmnetp: in `routers/rslogging.py` the `log_book_event` does just that.  As well as `routers/assessment.py` and the `get_assessment_results`

bnmnetp:
Regarding Core vs ORM -- I didn't change the queries at all.  I did change the inserts, but that was trivial.  

The only messing around I had to do was to get the async stuff right.  My understanding -- which seems to have played out -- is that 1.4 / 2.0 are unifying the Core vs ORM apis.

One thing I wanted to experiment with today was to see if the second `async with` block is really needed for the inserts.

Other than a day and a half of work I'm not highly invested in either Core or ORM.  But before we decide to switch back I think its worth looking at whether the automated mapping to Pydantic will even work with tables.

bjones1:
Sounds good. I'll write up the mapper and test on tables, then let you know.

bnmnetp:
The nested `async with session.begin()` seems like overkill.  The tests I have still pass without it and the db is populated correctly.

bookserver/session.py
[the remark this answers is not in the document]
bjones1: Thanks, I'll take a look.

bjones1 (commented):
Neat, thanks for checking. That's a good step forward toward simpler code!

bookserver/session.py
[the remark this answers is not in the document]
bjones1: I like the way these schemas work -- FastAPI is really nice. You did a great job finding a really solid framework!

bjones1:
I just pushed code with the auto Pydantic model generation from a Table, not a declarative base. (The code supports both styles, thought.) If you're fine, let's move to Core instead of ORM / declarative.

bjones1:
I'm getting a flake8 failure that `session` isn't defined in `assessment.py`. What's the correct fix?

bjones1:
Also, this PR is now pretty much what I want it to be -- a good first pass through the code. What other changes would you like to see in it?

When you approve, I can squash this down to a nicer set of commits.

bnmnetp:
You can comment out line 57 for now.... None of that code in the try block will work until I make further progress.   I don't know why flake8 isn't complaining for me....  It clearly should be.

So.. given that we do not need the double layer of `async with` AND that the query api is the same for both Core and ORM, and that I just converted it to ORM because that's what you had done with the Flask models, and how the examples of using async are written that way... what are your objections to ORM? I'm just not seeing that much difference and I want to be sure before I throw out the work I just did.

I think we should merge this PR before switching back if that is the final decision.

bjones1:
I now have mypy checks passing, plus fixed a fair amount of actual problems that mypy found,

bnmnetp:
Doing a little more research on ORM vs Core.... I want to make the best decision using the 1.4 / 2.0 future as our reference...  I'm certainly not a SQLAlchemy expert.

The select interface is unified and the preferred interface for 2.0 -- I think this is exactly what you want as it is the same as it was in Core for previous versions.  And again my queries that I have in this code which were written thinking we were going with Core did not need to be modified!  I think the orm way of doing queries was its biggest drawback and now that is gone.

I think the session.merge(Object) would save us a fair amount of work and be far less error prone in the insert-or-update use case.  You modify the object and call merge on it and commit it.  If the object does not exist (ie no id) then a new object is automatically created.

I think many of the issues with the ORM come from over use of the relationships or misunderstanding of relationships which I have always found to be quite ugly.  But this seems to be an area where we can do our own thing and not use that feature if we don't want to.  

I don't see us running into the implicit I/O scenario, but I've just read about it and maybe I'm not seeing it clearly.  Do you have a specific example from Runestone code where you see a risk of that?

Just a quick FYI -- The `async with session.begin():` context manger eliminates the need for the `async session.commit()` call because it is a context manager and does exactly what you would think a context manager would do in this case -- commit the transaction!

We will need to be much more mindful of doing commits with SQLAlchemy as we don't have the autocommit behavior of the tight integration with web2py and DAL which implicitly did commits after a successful api call.  I think this is another reason why I like isolating as much of the database functionality to crud.py as we can.

bjones1:
I am also not a SQLAlchemy expert -- I'm also reading about it. I've used it in the past, but much of that experience is not longer relevant. I'm also very excited about the new 1.4 / 2.0 approach where the Core and ORM are more unified. That's a big step forward.

Here are some rambling thoughts on databases with Python:

Design goals
==============
1. I want to stay as close to the underlying SQL as possible, since this is the language in which everything is actually implemented. Constructing difficult queries or debugging difficult problems always happens at the SQL level. Therefore, I want a Python interface that keeps me close as to the SQL as possible. Any time I spend translating from Python to SQL or SQL to Python is time wasted. The amount of time we've already spent trying to understand SQLAlchemy is just a foretaste of what we'll both be doing to make this work. The size of the docs is another hint of a lot of complexity to handle.
1. I want to write in some sort of vendor-neutral SQL. Here, I like what SQLAlchemy / DAL / etc. do -- I can define tables/indices/queries in one way, then run those on many different databases. (Practically, I want my queries to work on both sqlite and on Postgres.)
1. I want an async connection to the database. SQLAlchemy certainly shines in this regard.
1. I want to be able to compose queries programmatically, rather than gluing pieces of textual SQL together. I want everything automatically quoted to avoid SQL injection attacks. Again, most ORMs do well here.

So, my major design goal that's not clearly met with an ORM is to stay as close as possible to the underlying SQL. To me, the SQLAlchemy section on [Data Manipulation with the ORM](https://docs.sqlalchemy.org/en/14/tutorial/orm_data_manipulation.html#data-manipulation-with-the-orm) provides a great example of how these two are widely separated. For example, a select actually means an update followed by a select with a `RETURNING` clause to get the new IDs of object. Aack! When something goes wrong, debugging this will be quite difficult. Both PostgreSQL and SQLite offer upsert-like syntax; I want something that provides this at the Python level, instead of a large amount of complex Python library code that produces difficult-to-follow SQL.

I agree and am excited about the new SQLAlchemy features that use Python well -- for example, context managers which commit when closing are fantastic! I also agree that using relationships takes us farther from the underlying SQL. As you say, by not including these relationships in the tables, we can avoid this problem entirely. A plus of the SQLAlchemy ORM is that I prefer that style of defining the tables vs. the Core style.

So, I think I'm still looking for a mythical third option. However, of existing implementations, either SQLAlchemy Core or DAL are probably our best alternatives.

It might also be helpful to look at example of the common cases. Here's code I like:

``` Python
async def fetch_last_answer_table_entry(query_data: schemas.AssessmentRequest):
    assessment = EVENT2TABLE[query_data.event]
    tbl = answer_tables[assessment]
    query = (
        select(tbl)
        .where(
            and_(
                tbl.div_id == query_data.div_id,
                tbl.course_name == query_data.course,
                tbl.sid == query_data.sid,
            )
        )
        .order_by(tbl.timestamp.desc())
    )
        res = await session.execute(query)

    return res.scalars().first()
```

Here's a great example of using programmatic queries to compose a query with an obvious SQL outcome, a clear select statement, etc. In contrast, I'm not sure what SQL this generates:

``` Python
async def create_answer_table_entry(log_entry: schemas.LogItem):
    values = { # omitted
    }
    tbl = answer_tables[EVENT2TABLE[log_entry.event]]
    new_entry = tbl(**values)
        session.add(new_entry)
    return new_entry
```

I would prefer to write this as:
``` Python
async def create_answer_table_entry(log_entry: schemas.LogItem):
    values = { # omitted
    }
    tbl = answer_tables[EVENT2TABLE[log_entry.event]]
        session.exeute(insert(tb), [values])
```

An upsert would be a more interesting case to look at.

bnmnetp:
Thanks,

I think our disagreement may come down to a missing design goal and how much priority we place on it.  I want a tool like sqlalchemy to make me productive and reduce the amount of repetitive code I (and other less experienced developers) need to write.

So the example you liked is possible with ORM or Core and unless I'm missing something meets all four of your criteria as well as my additional.

I actually prefer the first version of `create_answer_table_entry` I know that that add does an insert of the new object into the database.  If I want to see the SQL I can turn on the flag that will spew that out.  But I think it is pretty apparent that you are creating a new object and adding it to the database.

For the upsert use case I prefer something like:

```
my_entry.value = newvalue
session.merge(my_entry)
```

This means I don't have to write a lot of error prone repetitive code that amounts to an `update my_table set my_entry = newvalue where id = whatever`. It also means I don't have to compose a query first to see if the thing is already in the database.

If you really want to do an explicit insert you can still do that with ORM just by using `tbl.__table__` I don't think that most of the time that would be a good idea, but it is there.

I'd like to resolve this so we can get moving again.  Any suggestions for resolving this decision?

bnmnetp:
One more quick note.  Just from an aesthetic point of view I really like that I don't have to say `tbl.c.div_id` everywhere.  I can just say `tbl.div_id` when using ORM.

bjones1:
Hmm. In the end, we can't really know how this will all end up, but just make our best educated guesses. Let's go with the ORM and see what happens.

bjones1:
I'd like to clean up this PR before merging it. What else would you like done here before the merge?

bnmnetp:
I think this is fine as it is.  We will continue to evolve.

I did think of one experiment I want to run that could tip the scales for me.  **Performance** if inserting new rows to useinfo and the xxx_answers tables is significantly slower in ORM than Core then that is a big factor.  

Our top three API calls are:

1. hsblog -- insert into useinfo and answers tables 
2. published -- lookup a user and a course to serve a book page
The top two are both double the next one
3. getAssessResults -- retrieve answers from xxx_answers table

I will check both ways and see how performance compares.

I also know the average response time on academy.

bjones1:
Sounds good!

bnmnetp:
Well, performance testing was pretty uninteresting....

I timed the insertion of 10,000 rows into the useinfo table.

I started by testing if the extra `async with` versus an explicit commit made any difference  in ORM mode.  And it did not.  I then defined the useinfo table in core mode and reran the test using `await session.execute(insert(logtable).values(**new_log))`. 

I then reran tests on a postgresql database and was mildly surprised when they failed immediately on a ForeignKey violation...  🤣 yeah, sqlite does not enforce those.  So postgresql was actually slightly slower in this test than sqlite!

I had kind of expected ORM to lose out because of the object creation.  But was surprised that there wasn't really any difference.  See notes below.

```
ORM - sqlite
For 10,000 on an empty useinfo table with context manager 0:00:58.647466
For 10,000 on an empty useinfo table with explicit commit 0:00:58.679958
ORM - postgres 
For 10,000 on empty useinfo table with explicit commit  0:01:03.433731 - foreign keys enforced
```

```
Core - sqlite
For 10,000 on an empty useinfo table with explicit commit 0:00:59.070297
Core-postgres
For 10,000 on an empty useinfo table with explicit commit 0:01:02.985901 Foreign keys enforced
```

bjones1:
Neat, good data. I'm glad to know there's not much performance difference.

bnmnetp (approved):

bnmnetp:
Is this ready to merge or were you still going to squash commits??

bjones1:
Let me do some squashing first...

bjones1:
OK, I squashed / cleaned the commits.

PR #11, opened 2021-04-26
WIP - move to use async sqlalchemy directly

bnmnetp:
ok @bjones1 This moves us to using async sqlalchemy directly.

Still more work to do.

1. Move rest of tables to ORM model 
2. Update the rest of the crud.py queries.
3. Write unit tests

BUT, it is working for basic login and page display.

There is a question in my mind as to whether we should be using FastAPI Depends for the async session as a parameter to (nearly) every endpoint or whether we should use create_session() as I have done for this work. I think I like the way I did it better, and if I'm understanding correctly sqlalchemy will use an underlying connection pool automatically.

In any case this gives us more to talk about.


bjones1:
Exciting! Let me do some more reading/thinking about this.

Also, when you have time, would you review https://github.com/bnmnetp/BookServer/wiki/Overall-documentation-structure?

bjones1:
I rebased my branch on your new async code, since I think that's definitely a good path forward.

bnmnetp:
I have looked at the wiki and I moved a few things around and added some on your comment at the end.  Its not clear how to just comment.

bjones1:
Thanks! The Wiki doesn't have good commenting abilities -- we can move this to Google Docs if you'd prefer. I appreciate the edits -- I wanted your input on the overall docs structure.

bnmnetp:
One thing I like about the CodeChat doc is that it has a clear delineation between the User and the Developer I think some organization along those lines would help this a lot too.

bnmnetp:
I think this is plenty of progress to show that the direct use of sqlalchemy is viable and we ought to merge this to main so we can experiment and make progress on other fronts.

I'm going to have to check to see why 3.8 is failing.  I'm using 3.9 for all dev work and pre_commit_check.py is passing all the time locally for me.

bjones1:
I'll leave a few comments, but I agree -- please merge this to the main branch. I'm stills struggling to understand sqlalchemy concepts, so I don't yet have an option on `create_session()` vs. FastAPI's dependency injection.

Helpful links: https://docs.sqlalchemy.org/en/14/orm/session_transaction.html#orm-session-vs-engine

PR #14, opened 2021-04-29
Misc fixes and TODO updates

bjones1:
It also fixes some confusion in the schemas and tests.

### DIFF
--- a/bookserver/crud.py
+++ b/bookserver/crud.py
@@ -81,15 +81,16 @@ async def create_answer_table_entry(log_entry: schemas.LogItem):
     values["timestamp"] = datetime.utcnow()
     rslogger.debug(f"hello from create at {values}")
     tbl = answer_tables[EVENT2TABLE[log_entry.event]]
+    # TODO: revalidate now that we know which table to validate against?
     new_entry = tbl(**values)
     async with async_session() as session:
         session.add(new_entry)
+        # TODO: remove this.
         await session.commit()
     rslogger.debug(f"returning {new_entry}")
     return new_entry
 
 
-# :index:`TODO`: **I think the idea here**, but the implementation will still need some special cases for getting the specific data for all the question types.
 async def fetch_last_answer_table_entry(query_data: schemas.AssessmentRequest):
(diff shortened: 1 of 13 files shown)


bookserver/crud.py
```diff
@@ -81,15 +81,16 @@ async def create_answer_table_entry(log_entry: schemas.LogItem):
     values["timestamp"] = datetime.utcnow()
     rslogger.debug(f"hello from create at {values}")
     tbl = answer_tables[EVENT2TABLE[log_entry.event]]
+    # TODO: revalidate now that we know which table to validate against?
     new_entry = tbl(**values)
     async with async_session() as session:
         session.add(new_entry)
+        # TODO: remove this.
```
bnmnetp, line 88: You need the `with async session.begin():` statement if you want to remove the explicit commit.

bnmnetp: I think I mentioned this before, but I find this mode of operation to be quite awkward.  You modify the code to essentially report a bug.  And I have to respond this way.  So that part of the conversation is in the history of the code but part of it is not.

I would prefer we use discussions or issues for this.

bjones1: I understand. I agree that it's awkward, but it's also a pain to find a problem in my editor, then find it in the github web interface, then file an issue. I'm moving toward this second model, though.

If I'm offline, I can only do the first model.

bjones1: I'll revise my PR to remove it.

bookserver/routers/books.py
```diff
@@ -103,11 +104,11 @@ async def serve_page(
     #   activity_info
     #   settings.google_ga
     await create_useinfo_entry(
```
bnmnetp, line 106: I'm not following this, and it appears to be a step back from using a well defined schema to a dictionary.  Can you give me some background on your thinking and the rationale for doing this?

bjones1: I'd like a chance to talk with you about this -- when are you free?

bjones1:
OK, this is ready now.

bnmnetp (approved):

PR #23, opened 2021-05-01
Use the SQLAlchemy Core for queries

bjones1:
In general, we don't want to allow ORM objects to "escape" from `crud.py`, so avoid the possibility of unintentional changes to these variables produces actual changes in the database.

One approach is to query using the ORM, then convert the results into a Pydantic struct. A simpler solution is to query the Core:
- The query syntax is very, very similar.
- The results from the query can be returned directly outside this module, automatically preventing ORM escapes.
- It's less error-prone (whoops! Forgot to wrap that ORM result before returning it...)
- It's probably a bit faster.

I found out (the hard way) that my initial solution used a Python 3.9-only feature. I also added `tox` to make it easier to test across multiple Python versions.

### DIFF
--- a/bookserver/crud.py
+++ b/bookserver/crud.py
@@ -16,9 +16,7 @@
 
 # Third-party imports
 # -------------------
-from .db import async_session
-
-# import sqlalchemy
+from .db import engine, async_session
 from sqlalchemy import and_
 from sqlalchemy.sql import select
 
@@ -27,7 +25,6 @@ from sqlalchemy.sql import select
 from .applogger import rslogger
 from . import schemas
 from .models import (
-    AuthUserValidator,
     Useinfo,
     AuthUser,
     Courses,
@@ -87,7 +84,7 @@ async def fetch_last_answer_table_entry(query_data: schemas.AssessmentRequest):
     assessment = EVENT2TABLE[query_data.event]
     tbl = answer_tables[assessment]
     query = (
-        select(tbl)
+        select(tbl.t)
         .where(
             and_(
                 tbl.div_id == query_data.div_id,
@@ -97,27 +94,23 @@ async def fetch_last_answer_table_entry(query_data: schemas.AssessmentRequest):
         )
         .order_by(tbl.timestamp.desc())
     )
-    async with async_session() as session:
-        res = await session.execute(query)
+    async with engine.connect() as conn:
+        res = await conn.execute(query)
         rslogger.debug(f"res = {res}")
 
-    return res.scalars().first()
+    return res.first()
 
 
 async def fetch_course(course_name: str):
-    query = select(Courses).where(Courses.course_name == course_name)
-    async with async_session() as session:
-        res = await session.execute(query)
-    # When selecting ORM entries it is useful to use the ``scalars`` method
-    # This modifies the result so that you are getting the ORM object
-    # instead of a Row object. `See <https://docs.sqlalchemy.org/en/14/orm/queryguide.html#selecting-orm-entities-and-attributes>`_
-    return res.scalars().first()
+    query = select(Courses.t).where(Courses.course_name == course_name)
+    async with engine.connect() as conn:
+        res = await conn.execute(query)
+    return res.first()
 
 
 async def fetch_user(user_name: str):
-    query = select(AuthUser).where(AuthUser.username == user_name)
-    async with async_session() as session:
-        res = await session.execute(query)
+    query = select(AuthUser.t).where(AuthUser.username == user_name)
+    async with engine.connect() as conn:
+        res = await conn.execute(query)
         rslogger.debug(f"res = {res}")
-    user = res.scalars().first()
-    return AuthUserValidator.from_orm(user) if user else None
+    return res.first()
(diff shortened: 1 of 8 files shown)


bookserver/crud.py
```diff
@@ -97,27 +94,23 @@ async def fetch_last_answer_table_entry(query_data: schemas.AssessmentRequest):
         )
         .order_by(tbl.timestamp.desc())
     )
-    async with async_session() as session:
```
bnmnetp, line 100: These queries were working just fine. I thought we had come to agreement on our Sqlalchemy strategy...  Why the changes??

bnmnetp:
If we are going to use tox can we then eliminate the `pre_commit_check.py` script?  Seems like we should not duplicate.

bookserver/crud.py
bjones1: The more I'm using SQLAlchemy v1.4, the more I realize that the choice isn't exclusive. My thoughts:
- Defining tables ORM style is cleaner and more readable. Let's continue using that.
- Using columns ORM style (`tbl.div_id`) is cleaner and more readable than columns Core style (`tbl.c.div_id`). Let's continue using that.
- Most DB operations (queries, inserts, etc.) are very, very similar. For these, let's pick the style that meets our design goals.

So, my main point is: we can freely mix and match ORM vs. Core, picking the style that's best for the situation. In the case of a query, the options are:
- ORM: We `select(tbl)`, then get the results in a class. Since it's an ORM class, we then need to convert it to something else (a Pydantic class is probably easiest) so that accidental changes won't affect the database.
- Core: we `select(tbl.t)`, then get the results in a class-like object we can directly return. This seems slightly easier to me.

bjones1:
We need to run the unit tests on multiple versions of Python. CI does this, but is slow when tracking down / fixing a bug; tox enables us to run these tests locally much faster.

The other checks (flake8, black, sphinx-build, and even mypy) shouldn't vary with the Python version. So, while they're still important, there's no point (IMHO) in running these tests on multiple Python versions.

So, both are useful (with differing purposes) and both should be kept.

bookserver/crud.py
bnmnetp: At this point I think we have made things needlessly complicated.  Defining the model with ORM syntax, and creating objects with ORM syntax, then adding a metaclass to get the underlying `.__table__ ` so that we can return a Row instead of a Pydantic object.  I think this kind of complexity makes it hard for others to join in the project.

I'm going to be honest, I love Python because it is expressive and lets me get stuff done quickly.  I feel like we are perilously close to turning Python into Java with all of this validation and type hinting, and linting...  Every time I try to do a commit and can't because flake8 says I have an unused import I cringe and say a bad word.  

But I want us to be consistent, and I want us to have code that is accessible to newcomers who want to contribute, and I want the bookserver to be fast and super reliable.

So... I'm trying bend and see if I can get along with type hints and whatnot.   I'm trying to see FastAPI as a model project.

`crud.py` is still a work in progress, so I appreciate the experimentation.  But I'm also feeling a lot of urgency to have a decision made and get on with things.

1. I think it would be simpler for us to clearly choose Core or ORM and not some hybrid.
2. I agree that we want to avoid returning ORM objects and having open sessions and all of the behind the scenes automatic update that could lead to hard to debug issues.
3. We do not embed a lot of logic into our model objects.  The only place we really do that is in the dashboard and the performance there is horrible.  Largely because tons of stuff that could be done with better queries is done with Python loops. 
4. Most of ajax.py is very basic CRUD type operations.  No complex business logic to embed there.
5. I like the idea of being able to declare the kind of object each crud function is going to return for the sake of mypy but also because vscode pylance has good integration with pydantic and can do autocomplete.  sqlalchemy rows are just rows -- no autocompletion.

So, maybe I'm saying Core is the best way to go, but that I would still prefer to convert crud results to a pydantic class.  I think mypy will catch your whooops forgot to convert scenario very quickly!

bnmnetp:
I kind of agree but only up to a point.  When I'm debugging it is even faster to simply run pytest, I don't want or need to do the other checks or rebuild all of our docs when I'm just trying to get a unit test to pass.  I'm ok with having tox or some other tool tell me that what I just did works fine for 3.9 but not 3.8 even after I've got it working on 3.9.

It seems like we have 3 places to worry about now for testing.  tox, pre_commit_script, and our github workflow. (I know it uses pre_commit_script -- but we still have to setup the version matrix and env vars and...) 
I'm going to do some reading on tox since I've never used it before.  As a totally naive user it seems like I should be able to say tox XXX where XXX is a task.  But maybe that is not the design.

bjones1:
I'd say that it's three options for testing: pre_commit_check to run before a commit; pytest for a quick check; tox if struggling with bugs on a specific Python version. Pick the tool you need for the task at hand.

bnmnetp:
Here is what I was getting at.  We can modify the tox.ini file as follows

```
[testenv:precommit]

deps = 
    flake8
    mypy
    black

commands =
    flake8 bookserver
    black --check bookserver
    mypy bookserver

[testenv:docs]
whitelist_externals = poetry

commands = 
    poetry run sphinx-build -E -a . _build
```

To run pytest: `tox`
To run the linters `tox -e precommit`
To build the docs `tox -e docs`

Just an idea and it uses the tools without having to maintain another python script.

bjones1:
Nice, I like it! Great idea. I'll submit this as a separate PR.

PR #31, opened 2021-05-03
Fix: Select config, dburl in settings.

bjones1:
This is related to testing -- I need an easy way in tests to find the dburl and related settings. This tidies that process up.


bookserver/config.py
```diff
@@ -26,6 +27,13 @@
 
 # Settings
 # ========
+# Define the possible bookserver configurations. The values assigned must  be strings, since Pydantic will match these with environment variables.
+class Config(Enum):
```
bnmnetp, line 31: I think config was an ill conceived name on my part. I think maybe we should use FAPIConfig or something...  We use `WEB2PY_CONFIG` for the web2py server.  Using config here risks confusion with the Config for the Pydantic models and is just generally overused.

bookserver/config.py
```diff
@@ -48,6 +56,15 @@ class Settings(BaseSettings):
     dev_dburl: str = "sqlite+aiosqlite:///./runestone_dev.db"
     test_dburl: str = "sqlite+aiosqlite:///./runestone_test.db"
 
+    # Determine the database URL based on the ``config`` and the dburls above.
+    @property
```
bnmnetp, line 60: Does this still let you set the DB URLS in the environment?

I think it is critical to be able to continue to do that. Especially when we move to postgres and have embedded passwords.

bookserver/config.py
bjones1: How about `BOOK_SRV_CONFIG`? I agree that config is too generic.

bookserver/config.py
bjones1: Yes -- the only unsettable item is `database_url`, which is set based on the other variables instead of being set directly.

bookserver/config.py
bnmnetp: I'm good with `BOOK_SRV_CONFIG`

bookserver/config.py
bnmnetp: Oh right, this is a net add of an item the other database url variables are still there.

bnmnetp (approved):

PR #32, opened 2021-05-03
authed logging

bnmnetp:


bnmnetp:
This PR does a few things (all related)

1. Adds the CourseInstructor model so we can verify when a user is an instructor
2. Moves the check for a user being logged in to middleware -- request.state.user will be None if user not logged in, otherwise it will contain the User object.
3. Adds a function is_instructor() to check if the user in the request object is an instructor for the course
4. Updates get_assessment_results to use the authenticated user or else it will return a 401 not authorized status

From the browser it looks good so far!  We'll need to figure out the login/logout stuff for the tests though.

bookserver/session.py
```diff
@@ -41,3 +43,15 @@ async def load_user(user_id: str):
...
     rslogger.debug(f"Going to fetch {user_id}")
     return await fetch_user(user_id)
+
+
+async def is_instructor(request: Request) -> bool:
+    user = request.state.user
+    if user is None:
+        raise HTTPException(401)
+    elif user.course_name in [
+        x.course_name for x in await fetch_instructor_courses(user.id)
```
bjones1, line 53: Wouldn't this be better as a part of the query, instead of getting all courses then searching the resulting list in Python?

bnmnetp: I knew you were going to say that. 😄 

My thinking here is that there are cases where we want all of the courses for an instructor, but it is a pretty small number for everyone but me and maybe one or two others.  So we are making and searching a list of less than six things almost all the time.

This makes for a more generally useful function.

bnmnetp: I guess the other alternative would be to have course_id as a secondary optional parameter for the function in crud. If provided we return zero or one objects and if its not provided we return all.  

I like the idea of a bunch of fairly generic calls in crud that do the CRUD things.  And I like the idea of isolating the database queries rather than spreading them everywhere.  But we'll have to see how it goes as we feel our way through the first bunch of API calls like this.

bookserver/routers/assessment.py
```diff
@@ -39,15 +39,20 @@
...
+
+    # if the user is not logged in an HTTP 401 will be returned.
+    # Otherwise if the user is an instructor then use the provided
+    # sid (it could be any student in the class) If none is provided then
+    # use the user objects username
+    if await is_instructor(request):
+        if not request_data.sid:
+            request_data.sid = request.state.user.username
+    else:
+        request_data.sid = request.state.user.username
```
bnmnetp, line 55: This works as is, but I wonder if we should be raising a 401 here in the case where the user is not an instructor AND is trying to send along a user id that is not theirs.  

The way it will always return a result for a logged in user, it will just ignore the sid field of the request_data parameter if the user is not an instructor.
If someone really is trying to spoof the system I like the idea of not giving them anything and the 401 says "caught ya"

bnmnetp:
@bjones1 This has a lot of good stuff in it now and would be good to get merged.

It makes all the usual base courses and a single user testuser1 with the password xxx

Note -- If you want to build a book for this server you should be using the bookserver branch of the Components on github.

bookserver/config.py
```diff
@@ -81,11 +80,16 @@ def database_url(self) -> str:
     web2py_path: str = str(Path(__file__).parents[2] / "web2py")
 
     # This is the private key web2py uses for hashing passwords.
-    @property  # type: ignore
-    @lru_cache
```
bjones1, line 85: Why remove this? Otherwise, every time the code asks for the private key, it will have to open the file and read-read the data.

bookserver/config.py
```diff
@@ -81,11 +80,16 @@ def database_url(self) -> str:
...
+    @property
     def web2py_private_key(self) -> str:
-        with open(self.web2py_path, encoding="utf-8") as f:
-            return f.read()
+        authfile = Path(self.web2py_path) / "private" / "auth.key"
```
bjones1, line 85: Oops! Thanks for this fix.

bookserver/crud.py
```diff
@@ -118,10 +127,118 @@ async def fetch_course(course_name: str) -> CoursesValidator:
     return CoursesValidator.from_orm(res.scalars().first())
 
 
+async def create_course(course_info: CoursesValidator) -> CoursesValidator:
+    new_course = Courses(**course_info.dict())
+    async with async_session() as session:
```
bjones1, line 132: Why not async_session.begin(), then omit the commit()?

bookserver/crud.py
```diff
@@ -118,10 +127,118 @@ async def fetch_course(course_name: str) -> CoursesValidator:
...
+
+async def create_user(user: AuthUserValidator) -> int:
+    """
+    The given user will have the password in plain text.  First we will hash
+    the password then add this user to the database.
+    """
+    new_user = AuthUser(**user.dict())
+    print(settings.web2py_private_key)
+    crypt = CRYPT(key=settings.web2py_private_key, salt=True)
+    new_user.password = str(crypt(user.password)[0])
```
bjones1, line 158: Does this work? The format is `alg$salt$hash`, and here I only see the hash being generated.

bookserver/crud.py
```diff
@@ -118,10 +127,118 @@ async def fetch_course(course_name: str) -> CoursesValidator:
...
+        res = await session.execute(query)
+
+    course_list = [
+        CourseInstructorValidator.from_orm(x) for x in res.scalars().fetchall()
+    ]
+    return course_list
+
+
+async def create_initial_courses_users():
+    if settings.drop_tables == "Yes":
```
bjones1, line 201: This is a bit more work, but I think it would be neat if courses was populated based on the subdirectories found in `setting.book_path` (unless that course entry already existed). Then, users could download a book to the appropriate location, restart the server, and it's ready to go!

bookserver/crud.py
```diff
@@ -118,10 +127,118 @@ async def fetch_course(course_name: str) -> CoursesValidator:
...
+
+        for c in BASE_COURSES:
+            new_course = CoursesValidator(
+                course_name=c,
+                base_course=c,
+                term_start_date=datetime.date(2000, 1, 1),
+            )
+            await create_course(new_course)
+        # make a user
+        await create_user(
```
bjones1, line 234: I'd suggest moving this into the test code.

bookserver/routers/assessment.py
bjones1: I agree!

bookserver/session.py
```diff
@@ -41,3 +43,13 @@ async def load_user(user_id: str):
     """
     rslogger.debug(f"Going to fetch {user_id}")
     return await fetch_user(user_id)
+
+
+async def is_instructor(request: Request) -> bool:
+    user = request.state.user
+    if user is None:
+        raise HTTPException(401)
+    elif len(await fetch_instructor_courses(user.id, user.course_id)) > 0:
```
bjones1, line 52: Shouldn't be check that the user is an instructor for the provided course?

bookserver/config.py
```diff
@@ -81,11 +80,16 @@ def database_url(self) -> str:
...
+        with open(authfile, encoding="utf-8") as f:
+            return f.read().strip()
+
+    # web2py_private_key: str = "sha512:16492eda-ba33-48d4-8748-98d9bbdf8d33"
+    # if you want to reinitialize your database set this to Yes
+    # All data in the database will be lost
+    drop_tables: str = "No"
```
bjones1, line 92: This seems like a testing-only setting -- it seems like a dangerous thing.

Why not make this a `bool`?

bookserver/crud.py
bnmnetp: wouldn't it have to be `async_session().begin()` ??

bookserver/crud.py
bnmnetp: yes it works.
the salt is part of the password that is stored in the database. 

the call to crypt returns the full thing: `alg$salt$hash$

bookserver/crud.py
bnmnetp: Interesting idea... I've been leaning the other way and wanting to make the book stuff more database driven and less reliant on the names of directories and the presence of conf.py files in things that are not really packages.

bookserver/crud.py
bnmnetp: So much development relies on interactive work.  This makes it possible to get a working environment quickly without running a bunch of manual commands to get things populated.

bookserver/session.py
bnmnetp: That is exactly what this is doing.

bookserver/config.py
bnmnetp: I wanted to make it hard to be true

bookserver/crud.py
bnmnetp: Actually I think they have to be nested in order for you to have a session variable.  In that case it is just a matter of style.

```
    async with session.begin():
        session.execute(....)
```

versus

```
    session.execute(...)
    await session.commit()
```

bookserver/crud.py
bnmnetp: I've been developing and testing logging in as testuser1 with no problems 😄

bookserver/crud.py
bnmnetp: It is kind of like the `rsmanage initdb`

bookserver/config.py
bjones1: Here's my rewrite that now works:

``` Python
    # This is the private key web2py uses for hashing passwords.
    @property  # type: ignore
    def web2py_private_key(self) -> str:
        @lru_cache
        def read_key():
            key_file = Path(self.web2py_path) / "applications/runestone/private/auth.key"
            with open(key_file, encoding="utf-8") as f:
                return f.read().strip()
        return read_key()
```

bookserver/config.py
bjones1: I inteded `webpy_path` to point to the root web2py folder, so I'm using a slightly different path (see previous comment).

bookserver/config.py
bnmnetp: Maybe the condition should be:

```
if settings.book_sever_config in ["development", "test"] and settings.drop_tables == "Yes":
```
This would rule out the accidental dropping of tables in a production environment.

bookserver/crud.py
bjones1: I poked around -- the magic "one-liner" syntax is:

``` Python
async with async_session() as session, session.begin():
```

We should use this syntax, since it implies a rollback if an exception is raised as well as a commit on success. See https://docs.sqlalchemy.org/en/14/orm/session_basics.html#framing-out-a-begin-commit-rollback-block.

bookserver/crud.py
bjones1: Great!

bookserver/crud.py
bjones1: What I'm hoping for is a simple end-user experience: unzip a book, put it in the right place, and it just works.  I'm concerned the database is less accessible than this process.

bookserver/session.py
bjones1: Oh, duh, I missed the second parameter.

bookserver/config.py
bjones1: I like it.

bookserver/crud.py
bnmnetp: I agree that we want to make it easy.  But the number of people that run their own servers is pretty small!  And... there is a new funded path for making custom books easy!  The github webhooks project.  I have at least partial funding for that through UT Austin.  So there will be a webpage that allows you to select a github repo and use that book for your course.

Can we agree that this should not hold up merging the PR?

bookserver/crud.py
bnmnetp: NICE!

bookserver/config.py
bnmnetp: I can see that, but I am currently in a mixed environment where web2py is in a Docker container and BookServer is not.  So pointing to the parent of the private directory makes sense for me.

bookserver/crud.py
bjones1: Yes, I agree.

bookserver/config.py
bnmnetp: this looks great, I'll include it in this PR

bjones1 (approved):

bookserver/crud.py
bnmnetp: OK, and also to your worries about changes to objects in the session, see: https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.close -- that seems to say that nad the end of the with block session.close() is called.

bjones1: I read a bit more -- since we're using a `sessionmaker`, even [better syntax](https://docs.sqlalchemy.org/en/14/orm/session_basics.html#using-a-sessionmaker) is `with Session.begin() as session` -- see below (copied from the docs):

``` Python
from sqlalchemy import create_engine
from sqlalchemy.orm import sessionmaker

# an Engine, which the Session will use for connection
# resources
engine = create_engine('postgresql://scott:tiger@localhost/')

# a sessionmaker(), also in the same scope as the engine
Session = sessionmaker(engine)

# we can now construct a Session() and include begin()/commit()/rollback()
# at once
with Session.begin() as session:
    session.add(some_object)
    session.add(some_other_object)
# commits the transaction, closes the session
```

The sqlalchemy docs do indicate that a `session.close()` (automatically performed by the `with` context manager) will detach everything from the session, which is exactly what we want. So, we can return results from `crud.py` as the original SQLAlchemy ORM classes if we want, instead of wrapping them in a Pydantic class.

bookserver/config.py
bjones1: I'll be using these assumptions about `web2py_path` in the test code -- let me know if that boogers something up for you.

PR #34, opened 2021-05-04
WIP: Progress toward Runestone Components tests.

bjones1:
This is a preview of my work so far -- expect force pushes, etc. as I progress.

Things to do:
- Need a way to log out of the server.
- Need a way to create a user
- Create a test user object. Perhaps wrap the existing auth_user object?
- Testing


bjones1:
I think this (sorta) works. You'll need to merge the updated bookserver branch on Runestone Components, assuming you're testing with SQLite. I haven't tested yet with PostgreSQL.

The tests come up then fail, since the components don't load yet. I've ported two tests, but the rest will be easy.

bookserver/crud.py
```diff
@@ -115,22 +115,21 @@ async def fetch_last_answer_table_entry(
 
 # Courses
 # -------
-async def fetch_course(course_name: str) -> CoursesValidator:
-    query = select(Courses).where(Courses.course_name == course_name)
+async def fetch_base_course(base_course: str) -> CoursesValidator:
```
bnmnetp, line 118: fetching a course is a necessary function.  not all courses are base courses.

bookserver/crud.py
```diff
@@ -115,22 +115,21 @@ async def fetch_last_answer_table_entry(
...
-    return CoursesValidator.from_orm(res.scalars().first())
+        # When selecting ORM entries it is useful to use the ``scalars`` method
+        # This modifies the result so that you are getting the ORM object
+        # instead of a Row object. `See <https://docs.sqlalchemy.org/en/14/orm/queryguide.html#selecting-orm-entities-and-attributes>`_
+        course = res.scalars().one_or_none()
+        return CoursesValidator.from_orm(course) if course else None
 
 
-async def create_course(course_info: CoursesValidator) -> CoursesValidator:
+async def create_course(course_info: CoursesValidator):
```
bnmnetp, line 129: There are many use cases where it is useful to get back the thing you just added.  Why return None?  At least maybe a boolean to indicate success / failure.

bookserver/db.py
```diff
@@ -52,6 +51,10 @@ async def init_models():
         await conn.run_sync(Base.metadata.create_all)
 
 
+async def term_models():
```
bnmnetp, line 54: Maybe some comments/documentation on what this is for?

bookserver/crud.py
bjones1: We should probably have another function -- `fetch_course` vs `fetch_base_course`. So far, we're only using `fetch_base_course`, so I just renamed it. I'd suggest postponing `fetch_course` until we have a need for it.

bookserver/crud.py
bjones1: For returning a boolean -- how do we want to handle exceptions/failures for crud operations? Trap the exception and return something nice, or allow the exception to terminate the operation? Or add another exception handler class for database exceptions that will return something besides a 501 plus log the error?

We could certainly return exactly what was passed to the function. I haven't needed that in my code, but I'm happy to provide it if you think it's useful.

bookserver/db.py
bjones1: I'll add docs. I spent a long time finding out that failing to terminate the engine prevents PostgreSQL from truncating / dropping tables (used to reset the database to a clean state before each test).

bookserver/crud.py
bnmnetp: Well, I knew the need was coming soon, so thats why I added the fetch_course when I wrote create.  Its early on with lots of things coming fast, so everything is not always going to end up perfectly in sync.

bookserver/crud.py
bnmnetp: The difference is that the returned value would have the id, which is what is needed in many cases.

One thing I do like about web2py is the tickets it generates. (well I don't like the raw form of the tickets) but the idea that whenever there is a crash it stores the crash in the errors folder so that I can investigate.

We could do that with an exception handler middleware that would also return the stack trace as part of the javascript response.  Rather than just a 501 return code.
over time this would allow our client side code to be more user friendly in providing the users the detail they need to report errors.

bjones1: I like the idea of middleware to capture exceptions and let us know!

I'll add returns of the created objects.

bnmnetp: Thanks, just as an example, logBookEvent does check the value returned from `create_useinfo_entry` to determine what to return to the client.  Its not a great example as I don't know that anything checks the value returned by logBookEvent in the browser, but still....

bnmnetp: My next PR will have everything in rslogger.py returning a `JSONResponse` object.

bjones1:
This is ready for review. It fails, but only because the server functionality isn't implemented yet. AFAIK, all the tests here will run as soon as we enable them; to keep the feedback / errors smaller, I've only enabled the test for multiple-choice questions.

When all these pass, we'll have pretty good coverage of much of the book server's functionality!

PR #35, opened 2021-05-05
Make timezone offset available in req.state

bnmnetp:


bookserver/routers/rslogging.py
```diff
@@ -71,3 +73,17 @@ async def log_book_event(entry: LogItemIncoming, request: Request):
...
+    response: Response, tzreq: TimezoneRequest, RS_info: Optional[str] = Cookie(None)
+):
+    if RS_info:
+        values = json.loads(RS_info)
+    else:
+        values = {}
+    values["tz_offset"] = tzreq.timezoneoffset
+    response.set_cookie(key="RS_info", value=str(json.dumps(values)))
+    rslogger.debug("setting timezone offset in session %s hours" % tzreq.timezoneoffset)
+    return "done"
```
bjones1, line 89: We might want to define some sort of standard JSON response protocol for out endpoints -- something to signal success or error, a standardize place to put additional data, etc.

bjones1: Perhaps a JsonResponse Pydantic class, which we inherit from to add additional fields?

bookserver/routers/rslogging.py
```diff
@@ -71,3 +73,17 @@ async def log_book_event(entry: LogItemIncoming, request: Request):
...
+
+@router.post("/set_tz_offset")
+def set_tz_offset(
+    response: Response, tzreq: TimezoneRequest, RS_info: Optional[str] = Cookie(None)
+):
+    if RS_info:
+        values = json.loads(RS_info)
+    else:
+        values = {}
+    values["tz_offset"] = tzreq.timezoneoffset
```
bjones1, line 86: I'd suggest a `constants.py` file or something like that, where we put repeated strings such as `RS_info` and `tz_offset`.

bookserver/routers/rslogging.py
bnmnetp: I think something like what we do with the exception handler would be good.

```
    return JSONResponse(
        status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
        content=jsonable_encoder({"detail": exc.errors()}),
    )
```

If the status code is not 200 then there should be the detail key in the json payload.
If the status code IS 200 then there should be a key "result" maybe that contains whatever response we have usually returned.

bjones1: If we use the status code to determine success (== 200 means success, != 200 is failure), then the payload of both could be some soft of result/detail. I'd like to structure / regularize this a bit. I'll think on it some...

bnmnetp: I'm just looking at a good case to think about right now.

In `runlog`  we can think about 4 "failure" modes:

1. The request comes from a not logged in user -- ignore the request
2. The request comes from a logged in user but the request.state.user.course name does not match the course name on the incoming data. -- ie the user changed courses in another tab
3. the incoming data says the user is logged in, but there is no request.state.user object -- the session expired
4. the user is all good but they are in a paired programing setup and their partner is invalid.

In the first 3 cases nothing will get saved to the database.

In the last case the logged in user's data is saved to the database but the "partner's" data is not saved.

PR #38, opened 2021-05-10
Fix: Use asyncio for test framework.

bjones1:


bookserver/db.py
```diff
@@ -30,7 +30,7 @@
 
 # TODO: Remove the ``echo=True`` when done debugging.
 engine = create_async_engine(
-    settings.database_url, connect_args=connect_args, echo=True
+    settings.database_url, connect_args=connect_args, #echo=True
```
bnmnetp, line 33: I suggest we make the echo parameter something we configure in settings.

bnmnetp (approved):

PR #41, opened 2021-05-12
Fix: Make from_orm(None) return None.

bjones1:


bjones1:
This is ready for review.

bjones1:
Oops, I accidentally pushed a couple other fixes to this branch (docs, remove LogItemIncoming). Let me know if you want me to back them out.

bookserver/crud.py
```diff
@@ -77,11 +77,11 @@ async def create_useinfo_entry(log_entry: UseinfoValidation) -> UseinfoValidatio
 # xxx_answers
 # -----------
 async def create_answer_table_entry(
-    # The correct type is one of the validators for an answer table; we use LogItemIncoming as a generalization of this.
```
bnmnetp, line 80: Why??  going to Type[Any] or Dict seems like a real step backward and does not do any conversion or validation.

bjones1: All the validation happens on line 88 below, when we constructor a validator for the correct event type. The previous code essentially checks the same values twice:

- Using the `LogItemIncoming` type annotation on line 81.
- Constructing the correct validator on line 88 below.

So, the entire LogItemIncoming type is both incorrect and redundant:
- Its types are as specific as the validator for the correct event type. What it claims as valid may not be.
- The table-specific validator will do a better job of checking the data.
- It's more code to maintain.

So, this PR DRYs the code by removing it.

bnmnetp: The unfortunate side effect is that the generated api documentation now has no idea and can give no guidance on what information is going to be in the payload.  I think that was a real benefit of this approach for both documentation and development purposes.

I guess we can still use the documentation interface for testing, but we won't have the benefit of a pydantic model to tell us what is expected.

bjones1: That's true -- we do lose some built-in docs. However, I think the benefit of removing redundant code outweighs the benefit of automatic endpoint docs. I also like that updates to the models will be immediately reflected in the validators, while they'll require a manual update of the LogItemIncoming (which might be overlooked, then cause failures/problems).

bjones1:
Per our conversation, I've eliminated the LogItemIncoming removal.

bnmnetp (approved):

PR #46, opened 2021-06-11
Support for user progress tracking

bnmnetp:


poetry.lock
```diff
@@ -21,7 +21,7 @@ typing_extensions = ">=3.7.2"
 name = "alabaster"
 version = "0.7.12"
 description = "A configurable sidebar-enabled Sphinx theme"
-category = "main"
+category = "dev"
```
bjones1, line 24: I would omit changes to this during development -- it just adds a bunch of noise to the commits.

bookserver/crud.py
```diff
@@ -35,13 +35,15 @@
     CodeValidator,
     CourseInstructor,
     CourseInstructorValidator,
+    UserSubChapterProgres,
```
bjones1, line 38: Should be spelled progress.

bookserver/models.py
```diff
@@ -559,6 +559,19 @@ class UserChapterProgres(Base, IdMixin):
     status = Column(Integer, nullable=False)
 
 
+class UserState(Base, IdMixin):
+    __tablename__ = "user_state"
+
+    user_id = Column(Integer, index=True)
```
bjones1, line 565: Is this a foreign key in some other table? If so, it should be marked that way.

All these should have NOT NULL constraints.

bookserver/routers/rslogging.py
```diff
@@ -179,3 +188,145 @@ async def runlog(request: Request, response: Response, data: LogRunIncoming):
...
+
+# completion tables
+# =================
+#
+# This section contains implementations of endpoints for tracking progress
+
+
+# updatelastpage
+# --------------
+@router.post("updatelastpage")
```
bjones1, line 201: It would be great to have a link to this endpoint in the RS JS.

bookserver/routers/rslogging.py
```diff
@@ -179,3 +188,145 @@ async def runlog(request: Request, response: Response, data: LogRunIncoming):
...
+async def updatelastpage(request: Request, request_data: LastPageIncoming):
+    if request_data.last_page_url is None:
+        return  # todo:  log request_data, request.args and request.env.path_info
+
+    lpd = LastPageData(**request_data.dict())
+
+    lpd.last_page_chapter = request_data.last_page_url.split("/")[-2]
```
bjones1, line 208: Some comments would be nice: "The URL provided must be ``chapter/subchapter.html``. Therefore, this code..."

bookserver/routers/rslogging.py
```diff
@@ -179,3 +188,145 @@ async def runlog(request: Request, response: Response, data: LogRunIncoming):
...
+    lpd.last_page_subchapter = ".".join(
+        request_data.last_page_url.split("/")[-1].split(".")[:-1]
+    )
+
+    if request.state.user:
```
bjones1, line 213: I'd suggest putting this at the beginning of the function. Without a user, this endpoint does nothing (as I understand it).

bookserver/schemas.py
```diff
@@ -154,3 +155,24 @@ class LogRunIncoming(BaseModelNone):
...
+# Schemas for Completion Data
+# ---------------------------
+class LastPageDataIncoming(BaseModel):
+    last_page_url: str  # = Field(None, alias="lastPageUrl") is the manual way
+    course_id: str = Field(None, alias="course")
+    completion_flag: str
+    last_page_scroll_location: str
+    # todo: this should really be an int
+
+    # We can automatically create the aliases!
```
bjones1, line 169: Very nice! A link to the Pydantic docs on this would be helpful.

bookserver/schemas.py
```diff
@@ -154,3 +155,24 @@ class LogRunIncoming(BaseModelNone):
     suffix: Optional[str]
     partner: Optional[str]
     sid: Optional[str]
+
+
+# Schemas for Completion Data
+# ---------------------------
+class LastPageDataIncoming(BaseModel):
+    last_page_url: str  # = Field(None, alias="lastPageUrl") is the manual way
+    course_id: str = Field(None, alias="course")
```
bjones1, line 164: Is this something we could instead change in the JS to make it consistent, instead of working around it here?

pyproject.toml
```diff
@@ -37,6 +37,7 @@ aiosqlite = "^0.17.0"
 
 # Development dependencies
 # ========================
+pyhumps = "^3.0.2"
```
bjones1, line 40: Put this above the title.

bookserver/crud.py
```diff
@@ -276,3 +278,39 @@ async def create_initial_courses_users():
                 course_id=12,
             )
         )
+
+
+async def update_user_state(user_data: schemas.LastPageData):
+    stmt = (
+        update(UserState)
+        .where(
+            (UserState.user_id == user_data.user.id)
```
bjones1, line 287: I don't understand this code -- where is `user` coming from in `user_data`?

bookserver/crud.py
```diff
@@ -276,3 +278,39 @@ async def create_initial_courses_users():
...
+                (UserSubChapterProgres.course_name == user_data.course_id)
+                | (
+                    UserSubChapterProgres.course_name == None
+                )  # Back fill for old entries without course
+            )
+        )
+        .values(**user_data.dict())
+    )
+    async with async_session() as session:
+        res = await session.execute(stmt)
```
bjones1, line 315: Does the update statement actually return something useful?

poetry.lock
bnmnetp: I'm not trying to change it!!

bookserver/crud.py
bnmnetp: Huh, the autogenerator tool seems to have dropped the second  s for some reason.

bookserver/routers/rslogging.py
bnmnetp: I appreciate the comments, but know that I am just copying in the code from the current implementation and trying to do the minimal amount of work to get it running as quickly as possible.  There is a tradeoff between perfection and all of the other projects on my plate.  I honestly don't have time to redesign every endpoint this summer, although I realize that now is an ideal time to do so.

That said, I am not sure this should even be an API endpoint anymore.  With pages dynamically generated we can update these tables from the serve_page function in books.py when the page request is made.  It will slow down page serving ever so slightly but it eliminates an api call as well!  Seems like a reasonable tradeoff to me... thoughts??

pyproject.toml
bnmnetp: I try to use poetry to add and remove things instead of manually editing the `pyproject.toml` file.  So I don't really think about updating it for documentation purposes.

bookserver/crud.py
bnmnetp: I think you just found a bug.

bjones1:
Is this ready for review? I wasn't sure where in the dev process you are with this one. I've provided some review, but need to look at it again when it's ready.

bnmnetp:
I think this is ready for more review.

Test coverage is not what I would expect given that this new code runs with every page load.  Still, I think I should add a couple of tests to make sure that stuff is working.... We do not have any selenium tests that look at the table of contents decorating or other student progress indicators.  Should we be doing all of our testing through selenium?  It seems redundant to write non-selenium unit tests for things that could also be more thoroughly tested with selenium....

bnmnetp:
I agree that using the "request review" feature is the clearest way to indicate things are ready for review.  Also having a PR as a draft and then promoting is also good practice.

bjones1:
Sounds good!

bnmnetp:
My biggest remaining confusion is the coverage report.  It is telling me that rslogging.py is only at 43% when it should be closer to 90%.  It seems that endpoints invoked from the selenium tests I wrote are not getting counted in test coverage...

bjones1:
I wonder if the way that Uvicorn launches an ASGI app (in another process? I don't know) prevents coverage from monitoring the code's executing. I Googled a bit, but didn't find anything obvious.

bnmnetp:
I think it is because you are starting a different server...

```
with TestClient(app) as client:
    foo
```
Starts a server for testing.

bookserver/routers/assessment.py
```diff
@@ -76,3 +79,68 @@ async def get_assessment_results(
...
+              "history": [code, code, code],
+              "timestamps": [ts, ts, ts]
+            }
+    """
+    acid = request_data.acid
+    sid = request_data.sid
+    # if request_data.sid then we know this is being called from the grading interface
+    # so verify that the actual user is an instructor.
+    if sid:
+        if request.state.user and request.state.user.username != sid:
```
bjones1, line 120: Won't this be a problem if the instructor wants to look at their own grades?

bookserver/routers/assessment.py
```diff
@@ -76,3 +79,68 @@ async def get_assessment_results(
...
+    # In this case, the request is simply from a student, so we will use
+    # their logged in username
+    elif request.state.user:
+        sid = request.state.user.username
+        course_id = request.state.user.course_id
+    else:
+        raise HTTPException(401)
+
+    res: Dict[str, Any] = {}
+    if sid:
```
bjones1, line 136: Will this ever be false?

bjones1 (changes requested):
This is a huge PR, so my review so far only covers about half of it. I'll work through more code as I have time.

bookserver/routers/books.py
```diff
@@ -57,20 +57,38 @@
...
     )
     rslogger.debug(f"GETTING: {filepath}")
-    return FileResponse(filepath)
+    if os.path.exists(filepath):
+        return FileResponse(filepath)
+    else:
+        raise HTTPException(404)
 
 
 @router.get("/published/{course:str}/_images/{filepath:path}")
```
bjones1, line 76: This seems like duplicate code -- can you chain the two decorators on the same function instead?

bookserver/routers/books.py
```diff
@@ -89,6 +107,11 @@ async def serve_page(
 ):
     rslogger.debug(f"user = {user}, course name = {course_name}")
     # Make sure this course exists, and look up its base course.
+    if user:
```
bjones1, line 110: Add a note somewhere that these values are being passed to JavaScript, hence `"true"` instead of `True` and so on (ick).

bookserver/routers/rslogging.py
```diff
@@ -149,7 +165,7 @@ async def runlog(request: Request, response: Response, data: LogRunIncoming):
     await create_useinfo_entry(UseinfoValidation(**useinfo_dict))
 
     # Now add an entry to the code table
```
bjones1, line 167: Comment that the code table used `acid` instead of `div_id`.

bookserver/routers/rslogging.py
```diff
@@ -179,3 +195,135 @@ async def runlog(request: Request, response: Response, data: LogRunIncoming):
...
+# =================
+#
+# This section contains implementations of endpoints for tracking progress
+
+
+# updatelastpage
+# --------------
+# see :ref:`processPageState`
+@router.post("/updatelastpage")
+async def updatelastpage(request: Request, request_data: LastPageDataIncoming):
```
bjones1, line 210: Add a hyperlink to the JS that uses this endpoint.

bookserver/routers/rslogging.py
```diff
@@ -179,3 +195,135 @@ async def runlog(request: Request, response: Response, data: LogRunIncoming):
...
+    if request_data.last_page_url is None:
+        return  # todo:  log request_data, request.args and request.env.path_info
+    if request.state.user:
+        lpd = request_data.dict()
+        rslogger.debug(f"{lpd=}")
+
+        lpd["last_page_chapter"] = request_data.last_page_url.split("/")[-2]
```
bjones1, line 217: Document this nasty stuff. Ideally, hyperlink to the same hackery in the Runestone Components. Ick.

bookserver/routers/rslogging.py
```diff
@@ -179,3 +195,135 @@ async def runlog(request: Request, response: Response, data: LogRunIncoming):
...
+        lpd["last_page_subchapter"] = ".".join(
+            request_data.last_page_url.split("/")[-1].split(".")[:-1]
+        )
+        lpd["last_page_accessed_on"] = datetime.utcnow()
+        lpd["user_id"] = request.state.user.id
+
+        lpdo: LastPageData = LastPageData(**lpd)
+        await update_user_state(lpdo)
+        await update_sub_chapter_progress(lpdo)
+        # The components don't ever look at a result from this
```
bjones1, line 227: I agree; we should also check the returned value from endpoints in the JS for error / success and inform the user.

bookserver/routers/rslogging.py
```diff
@@ -179,3 +195,135 @@ async def runlog(request: Request, response: Response, data: LogRunIncoming):
...
+        # to be ported somewhere....
+    return make_json_response(detail="Success")
+
+
+# _getCompletionStatus
+# --------------------
+@router.get("/getCompletionStatus")
+async def getCompletionStatus(request: Request, lastPageUrl: str):
+    if request.state.user:
+        last_page_chapter = lastPageUrl.split("/")[-2]
```
bjones1, line 244: DRY this (same stuff used earlier)

bookserver/schemas.py
```diff
@@ -17,6 +17,7 @@
 # Third-party imports
 # -------------------
 from pydantic import BaseModel, BaseConfig, create_model, constr, validator, Field
+from humps import camelize  # type: ignore
```
bjones1, line 20: Why the `# type: ignore` here?

bjones1 (changes requested):

bjones1:
OK, I've made a full pass through this PR and provided feedback.

bjones1:
Another to-do item -- more test coverage. I'd suggest merging the coverage fix then seeing what needs tests.

bookserver/schemas.py
bnmnetp: Because mypy was being a pain and claiming it was not used.

bookserver/routers/rslogging.py
bnmnetp: Sometimes you just gotta parse a string man

bnmnetp: Added comments

bookserver/routers/assessment.py
bnmnetp: I guess not, now that we are using the HTTPException inthe else clause.

bookserver/crud.py
```diff
@@ -220,6 +229,24 @@ async def create_code_entry(data: CodeValidator):
...
 
 
+async def fetch_code(sid: str, acid: str, course_id: int):
+    query = (
+        select(Code)
+        .where(
+            (Code.sid == sid)
+            & (Code.acid == acid)
+            & (Code.course_id == course_id)
+            & (Code.timestamp != None)  # noqa: E711
```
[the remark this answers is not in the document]
bnmnetp: It is probably a leftover condition to workaround a data problem in the database when there WERE rows that had a null value for timestamp.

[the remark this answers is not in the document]
bnmnetp: the current database has no such rows so it seems redundant at this point.

bookserver/crud.py
```diff
@@ -274,5 +306,170 @@ async def create_initial_courses_users():
...
+    )
+    async with async_session.begin() as session:
+        res = await session.execute(stmt)
+    rslogger.debug("SUCCESS")
+    return res
+
+
+async def update_sub_chapter_progress(user_data: schemas.LastPageData):
+    ud = user_data.dict()
+    ud.pop("last_page_url")
```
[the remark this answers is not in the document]
bnmnetp: It takes a LastPageData object It has to get changed somewhere so why not here?

bookserver/routers/rslogging.py
bjones1: I like this tradeoff -- remove an endpoint, plus remove JS that calls the endpoint and does processing, and put all the code in one place. One of the tough things about client/server development is the necessity of (effectively) scattering a single unit of code across both client and server. I'm all for putting it in one place.

bookserver/routers/rslogging.py
bjones1: :) thanks!

bookserver/crud.py
```diff
@@ -274,5 +306,170 @@ async def create_initial_courses_users():
...
+        )
+        .order_by(UserState.last_page_accessed_on.desc())
+    )
+
+    async with async_session() as session:
+        res = await session.execute(query)
+        # for A query like this one with columns from multiple tables
+        # res.first() returns a tuple
+        rslogger.debug(f"LP {res}")
+        PageData = namedtuple("PageData", [col for col in res.keys()])  # type: ignore
```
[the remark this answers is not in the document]
bnmnetp: No when you are joining across classes it does not.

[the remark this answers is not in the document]
bnmnetp: No, if it did I would have used it.

bnmnetp:
I know this is a big PR, but it opened up a number of cans of worms that we hadn't explored with the previous work.  I think I've addressed most everything.  I think a few things are a matter of style and priority.  I tend to just get something going as I'm feeling an urgent need to get this book server up and running, rather than polishing and perfecting.

So far, I have to admit that I think mypy is more of a pain than its worth.  I end up frustrated and feeling like I waste a lot of time just to figure out how to declare something in a way that will make mypy happy.

bnmnetp:
I'm going to go ahead and merge this so I can unblock some other stuff.

PR #48, opened 2021-06-18
Alembic / migration

bjones1:
This WIP shows progress toward migrating PostgreSQL from web2py to the BookServer. Notes:

- I've updated many of the indices based on what's created by rsmanage. Is this the correct approach? Or were the indices as previous written better?
- I've hand-written a script to rename all indicates created by rsmanage to the BookServer naming convention; see `rename_indices.sql`. It also includes the beginning of fixing NULL records where the new constraints specify NOT NULL.
- After running this script then creating a migration (use `alembic revision -m "testing" --autogenerate`), the resulting Python file has a LOT of `drop_index` statements. What do you see for your DB? Do we need to add more indices to the BookServer, or is what we have OK?
- I wrote a function to report on NOT NULL constraint violations. How many violations does your DB report? Mine has 59. I'd like to talk about where we should remove the NOT NULL and where we should backfill with reasonable defaults.


bjones1:
I've written all the migration my database needs (minus a few pending questions); @bnmnetp, try this on your test server and see what it reports.

bjones1:
This needs https://github.com/RunestoneInteractive/RunestoneComponents/pull/1213 to be merged in order to pass.

bjones1:
@bnmnetp, this is rebased to fix merge conflicts and ready to review.

bjones1:
@bnmnetp, pinging...

PR #54, opened 2021-06-23
Debug mode

bjones1:
Please review and merge #52 first, since this PR builds on it.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -76,6 +76,13 @@ def pytest_addoption(parser):
         help="Skip initialization of the test database.",
     )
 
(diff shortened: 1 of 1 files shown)


test/conftest.py
```diff
@@ -76,6 +76,13 @@ def pytest_addoption(parser):
         help="Skip initialization of the test database.",
     )
 
+    # This runs the server in a separate window with a usable console. A developer can add ``import pdb; pdb.set_trace()`` at any point in the bookserver to invoke the debugger and understand what's happening. The same technique also works well in the tests -- stop at a certain point in the test and (if running a Selenium-based test) look at the JavaScript console, or examine local variables in Python, etc.
+    parser.addoption(
```
bnmnetp, line 80: Is this an option that we pass to pytest when we run it?

bnmnetp (approved):

test/conftest.py
bjones1: Yes, exactly. It probably needs changes on Unix -- probably prefix the command with `xterm`?

PR #60, opened 2021-07-13
More assessment

bnmnetp:
This PR adds support for

* polls
* fill in the blank compare me functionality
* mchoice compare me functionality
* selectquestion functionality

It is a lot of new code in crud.py and assessment.py

Its nice to finally see the crud functions approach finally start to pay off.  In several places was able to make use of existing crud functions rather than re-porting queries time after time.

In some cases the return type has been made a list of strings as it saves the repetition of grabbing the string from a column after the function is called.


bjones1:
@bnmnetp, I assume this isn't ready yet since tests are still failing. Just let me know when you'd like my feedback.

bookserver/crud.py
```diff
@@ -84,6 +97,59 @@ async def create_useinfo_entry(log_entry: UseinfoValidation) -> UseinfoValidatio
...
+    )
+
+    async with async_session() as session:
+        rows = await session.execute(
+            query, params=dict(div_id=div_id, course_name=course_name)
+        )
+        return rows.all()
+
+
+async def fetch_top10_fitb(dbcourse, div_id):
```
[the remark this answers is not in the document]
bjones1: Document the return value -- I think it's a sequence of length 10 or less containing `answer, count` of the most common answers for the provided question, considering only questions after the provided date.

bookserver/crud.py
```diff
@@ -124,6 +190,21 @@ async def fetch_last_answer_table_entry(
         return rcd.validator.from_orm(res.scalars().first())  # type: ignore
 
 
+async def fetch_last_poll_response(sid: str, course_name: str, poll_id: str):
```
bjones1, line 193: I assume `poll_id` is the `div_id` of a poll question? That would be helpful to document.

bjones1: Also, `sid` is the same as `username`, except it also works for the unique identifier generated for students that aren't logged in?

bjones1: Document the return type; I think it's a string containing the last poll response to the given question.

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+        selected_id=selected_id,
+        points=points,
+        competency=competency,
+    )
+    async with async_session.begin() as session:
+        session.add(new_sqv)
+    return SelectedQuestionValidator.from_orm(new_sqv)
+
+
+async def fetch_selected_question(sid: str, selector_id) -> SelectedQuestionValidator:
```
bjones1, line 609: Missing type for `selector_id` -- I assume `str`?

bjones1: Also, I need docs somewhere (perhaps in models) on what/how selectquestions work. What is the `selector_id` (I think this is the `div_id` of the selectquestion) versus the `selected_id` (the `div_id` of the question to select? But don't we allow multiple ids, while this is a single id?)

bookserver/routers/rslogging.py
```diff
@@ -101,20 +100,32 @@ async def log_book_event(entry: LogItemIncoming, request: Request):
...
-    event = entry.event
-    if event in EVENT2TABLE:
-        rcd = runestone_component_dict[EVENT2TABLE[event]]
-        # Validate the fields for this component.
+    if entry.event in EVENT2TABLE:
+        rcd = runestone_component_dict[EVENT2TABLE[entry.event]]
+        if entry.event == "unittest":
+            # info we need looks like: "act":"percent:100.0:passed:2:failed:0"
+            ppf = entry.act.split(":")
+            entry.passed = int(ppf[3])
```
bjones1, line 109: Need to handle problems if these values don't parse -- return a 404 instead of just raising an exception.

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+        await session.execute(stmt)
+    rslogger.debug("SUCCESS")
+
+
+# Questions and Assignments
+# -------------------------
+
+
+async def fetch_question(
+    name: str, basecourse: Optional[str] = None
```
bjones1, line 639: Docs: is the `name` the same as the `div_id`?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+    where_clause = Question.name == name
+    if basecourse:
+        where_clause = where_clause & (Question.base_course == basecourse)
+
+    query = select(Question).where(where_clause)
+
+    async with async_session() as session:
+        res = await session.execute(query)
+        rslogger.debug(f"{res=}")
+        return QuestionValidator.from_orm(res.scalars().first())
```
bjones1, line 651: Should we do any checking if there are multiple values here?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+
+
+auto_gradable_q = [
```
bjones1, line 654: From the old server,

``` python
AUTOGRADEABLE = set(
    [
        "clickablearea",
        "dragndrop",
        "fillintheblank",
        "khanex",
        "mchoice",
        "parsonsprob",
        "quizly",
        "selectquestion",
    ]
)
```

LP is also autogradable. What about activecode?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+    "parsonsprob",
+    "dragndrop",
+    "fillintheblank",
+]
+
+
+async def fetch_matching_questions(request_data: schemas.SelectQRequest) -> List[str]:
+
+    if request_data.questions:
+        questionlist = request_data.questions.split(",")
```
bjones1, line 666: Aack, no JSON encoding for the data sent? Seems fragile.

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+auto_gradable_q = [
+    "clickablearea",
+    "mchoice",
+    "parsonsprob",
+    "dragndrop",
+    "fillintheblank",
+]
+
+
+async def fetch_matching_questions(request_data: schemas.SelectQRequest) -> List[str]:
```
bjones1, line 663: Docs: the return value is a list of `Question.name` from the db?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+
+        async with async_session() as session:
+            res = await session.execute(query)
+            rslogger.debug(f"{res=}")
+            questionlist = [q.name for q in res]
+
+    return questionlist
+
+
+async def fetch_assignment_question(assignment_name: str, question_name: str):
```
bjones1, line 699: Nitpick: return type.

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+        (Assignment.name == assignment_name)
+        & (Assignment.id == AssignmentQuestion.assignment_id)
+        & (AssignmentQuestion.question_id == Question.id)
+        & (Question.name == question_name)
+    )
+
+    async with async_session() as session:
+        res = await session.execute(query)
+        rslogger.debug(f"{res=}")
+        return AssignmentQuestionValidator.from_orm(res.scalars().first())
```
bjones1, line 711: Use `one_or_none` here instead of `first`?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+
+
+async def fetch_user_experiment(sid: str, ab_name: str) -> int:
```
bjones1, line 714: Docs: the returned `int` is the experiment group?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+
+
+async def create_user_experiment_entry(sid: str, ab: str, group: int):
+    new_ue = UserExperiment(sid=sid, exp_group=group, experiment_id=ab)
+    async with async_session.begin() as session:
+        session.add(new_ue)
+    return UserExperimentValidator.from_orm(new_ue)
+
+
+async def fetch_viewed_questions(sid, questionlist) -> List[str]:
```
bjones1, line 734: Nitpick: missing types

bjones1: Docs: the return value is a list of div_ids for questions that were viewed? Why does the existence of a sid entry imply they were viewed?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+        (Useinfo.sid == sid) & (Useinfo.div_id.in_(questionlist))
+    )
+    async with async_session() as session:
+        res = await session.execute(query)
+        rslogger.debug(f"{res=}")
+        rlist = [row.div_id for row in res]
+    return rlist
+
+
+async def fetch_previous_selections(sid) -> List[str]:
```
bjones1, line 745: Nitpick: missing type for `sid`.

bjones1: Docs: return value is a list of div_ids, correct?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+    sid: str, exam_id: str, course_name: str
+) -> TimedExamValidator:
+    query = (
+        select(TimedExam)
+        .where(
+            (TimedExam.div_id == exam_id)
+            & (TimedExam.sid == sid)
+            & (TimedExam.course_name == course_name)
+        )
+        .order_by(TimedExam.id.desc())
```
bjones1, line 763: This seems like a fallible approach --can we order by date instead?

bookserver/crud.py
```diff
@@ -494,3 +580,189 @@ async def create_user_chapter_progress_entry(
...
+    )
+    async with async_session() as session:
+        res = await session.execute(query)
+        rslogger.debug(f"{res=}")
+        return TimedExamValidator.from_orm(res.scalars().first())
```
bjones1, line 768: Use `one_or_none`?

bjones1 (changes requested):
One overall comment: I'm getting lost in the many functions we have. It would help me to have more docs of what a function does, what the parameters mean, and what the return type is as a first step.

bookserver/crud.py
bnmnetp: timestamps are too often identical the id has proven to be more reliable over time.

bookserver/crud.py
```diff
@@ -12,49 +12,61 @@
 #
 # Standard library
 # ----------------
+import datetime
```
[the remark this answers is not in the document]
bnmnetp: This is a problem with using the auto import tool in VSCode.  I have to remember to go back and order things.  it also puts things in the wrong place sometimes as it does not understand  your comments.

bookserver/crud.py
```diff
@@ -84,6 +97,59 @@ async def create_useinfo_entry(log_entry: UseinfoValidation) -> UseinfoValidatio
...
 
+async def count_useinfo_for(
+    div_id: str, course_name: str, start_date: datetime.datetime
+):
+    query = (
+        select(Useinfo.act, func.count(Useinfo.act).label("count"))
+        .where(
+            (Useinfo.div_id == div_id)
+            & (Useinfo.course_id == course_name)
+            & (Useinfo.timestamp > start_date)
```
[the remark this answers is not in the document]
bnmnetp: Not sure, just doing what we have been doing all along.

bookserver/crud.py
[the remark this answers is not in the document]
bjones1: Sounds good. I've seen a tool call [isort](https://github.com/PyCQA/isort), but not really evaluated it. Sorting manually for now works for me.

bookserver/crud.py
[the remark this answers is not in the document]
bjones1: Sounds good. Let's just document current behavior well and call that good.

bookserver/crud.py
bnmnetp: Yes but name isn't even uesd in this function.

bookserver/crud.py
bnmnetp: Activecodes are not autogradable unless they have unit tests.

I don't know enough abuot LP questions to know whether they should be in this list or not.

bnmnetp: I don't understand why this comment is here it doesn't seem to fit.

bookserver/crud.py
bnmnetp: again, ids have proven to be more reliable that timestamps as the timestamps are only accurate to a second.

bookserver/crud.py
bnmnetp: I don't understand the comment.  request_data is a schema validator object.  what more do you want?

bookserver/routers/rslogging.py
bnmnetp: I guess, but this doesn't currently happen as there is only one piece of code that generates these messages and it does so correctly.

bookserver/crud.py
bjones1: I see. All LP questions must have an associated unit test, so yes -- they are all autogradable.

bookserver/crud.py
bjones1: I mean the JS that sends the request data would ideally encode it as JSON instead of joining values with a comma. This is more a note for what needs fixing in the future.

bookserver/crud.py
bjones1: OK, sounds good.

bookserver/routers/rslogging.py
bjones1: Sounds good. Would you mind adding a link to the code that generates these values in the docs?

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+async def getaggregateresults(request: Request, div_id: str, course_name: str):
+    question = div_id
+
+    if not request.state.user:
+        return make_json_response(
+            status=status.HTTP_401_UNAUTHORIZED,
+            detail=dict(answerDict={}, misc={}, emess="You must be logged in"),
+        )
+
+    if course_name in (
```
bjones1, line 182: Why do all these courses start 90 days ago? I don't understand.

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
     ]
 
     return make_json_response(detail=res)
+
+
+# Used by :ref:`compareAnswers`
+@router.get("/getaggregateresults")
+async def getaggregateresults(request: Request, div_id: str, course_name: str):
```
bjones1, line 173: Nitpick: add return type.

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+    question = div_id
+
+    if not request.state.user:
```
bjones1, line 176: If a user isn't logged in, then `request.state` is None, so this causes an exception. I'd suggest adding a helper func to check this, since I see it a lot in the code.

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
     ]
 
     return make_json_response(detail=res)
+
+
+# Used by :ref:`compareAnswers`
+@router.get("/getaggregateresults")
```
bjones1, line 172: Docs would be helpful -- what's the purpose of this endpoint? The link is definitely nice, but doesn't appear to be pushed to the RS components repo yet.

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+
+    if course_name in (
+        "thinkcspy",
+        "pythonds",
+        "fopp",
+        "csawesome",
+        "apcsareview",
+        "StudentCSP",
+    ):
+        start_date = datetime.datetime.utcnow() - datetime.timedelta(days=90)
```
bjones1, line 190: `course` isn't set in this branch, but it's used later in the routine.

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+    else:
+        course = await fetch_course(course_name)
+        start_date = course.term_start_date
+
+    result = await count_useinfo_for(question, course_name, start_date)
+
+    tdata = {}
+    tot = 0.0
+    for row in result:
+        tdata[row[0]] = row[1]
```
bjones1, line 200: Docs would be nice -- what is `row[0]` and `row[1]`? Perhap use `for (act, count) in result`?

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+        tot += row[1]
+
+    tot = float(tot)
```
bjones1, line 203: Nitpick -- I don't think this is necessary, since `tot = 0.0` initially.

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+    rdata: Dict[str, float] = {}
+    miscdata = {}
+    correct = ""
+    if tot > 0:
+        for key in tdata:
+            all_a = key.split(":")
```
bjones1, line 209: More docs would be helpful -- I don't understand what the values are, why they're being split this way, etc. Perhaps just giving a typical value for the key would be enough.

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+            try:
+                answer = all_a[1]
+                if "correct" in key:
```
bjones1, line 212: I'm totally lost trying to understand what this code does. :(

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+
+@router.get("/getpollresults")
+async def getpollresults(request: Request, course: str, div_id: str):
+
+    # fetch summary of poll answers
+    result = await fetch_poll_summary(div_id, course)
+
+    opt_counts = {}
+
+    for row in result:
```
bjones1, line 246: More cryptic values below to my less-knowledgeable eyes -- again, something like `for (foo, bar) in result:` would help.

bookserver/routers/assessment.py
```diff
@@ -143,3 +166,355 @@ async def get_history(request: Request, request_data: HistoryRequest):
...
+    # opt_list holds the option numbers from smallest to largest
+    # count_list[i] holds the count of responses that chose option i
+    opt_list = sorted(opt_counts.keys())
+    count_list = []
+    for i in opt_list:
+        count_list.append(opt_counts[i])
+
+    total = sum(opt_counts.values())
+    user_res = None
+    if request.state.user:
```
bjones1, line 264: This seems a bit more complicated than necessary -- I would use:
``` Python
    if request.state.user:
        user_res = await fetch_last_poll_response(
            request.state.user.username, course, div_id
        )
        my_vote = int(user_res)
    else:
        my_vote = -1
```

bjones1 (changes requested):
I really like the comment you've written so far -- they are very helpful to me to understand the code. Thanks!

bookserver/crud.py
bnmnetp: I thought that you had this fixed so that .first() would work and the validator would be None in the case where there were no results.  In fact I'm pretty sure I tested and experimented with both approaches until I settled on this one as the working model.  Did that now change without me knowing?

bookserver/routers/assessment.py
bnmnetp: We don't have return types on any of the API endpoints as they all end up with some kind of JSON Response that I wouldn't even know how to specify.  I think its fine to leave them all like they are.

bookserver/routers/assessment.py
bnmnetp: `request.state` is always there.  If a user is not logged in the the .user will be None.

bookserver/routers/assessment.py
bnmnetp: These are all useful comments but I struggle with taking code that has worked for years and redoing NOW just for to get past your code review.  I have lots of other tasks on my plate.

bjones1: I understand -- time is a precious resource. I'm approving this PR and will leave all this to your discretion.

bjones1 (approved):

PR #66, opened 2021-08-12
Improve logging

bjones1:
This will help us more easily control how much info we want from a test session.

**WARNING** I set the default log level to `WARNING` instead of `INFO` to suppress most log noise by default. But there's always the chance there was something important in those details. Use `pytest --log-cli-level=INFO` to see lots more stuff.

### DIFF
--- a/bookserver/db.py
+++ b/bookserver/db.py
@@ -9,8 +9,8 @@
 #
 # Standard library
 # ----------------
-import logging
-
+# None.
+#
 # Third-party imports
 # -------------------
 # Use asyncio for SQLAlchemy -- see `SQLAlchemy Asynchronous I/O (asyncio) <https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html>`_.
@@ -22,9 +22,8 @@ from sqlalchemy.sql import select
 # Local application imports
 # -------------------------
 from .config import settings, BookServerConfig, DatabaseType
+from .applogger import rslogger
 
-
-logger = logging.getLogger(__name__)
 
 if settings.database_type == DatabaseType.SQLite:
     connect_args = {"check_same_thread": False}
@@ -60,7 +59,7 @@ async def init_models():
 
 # Look for any records that violate non-null constraints.
 async def check_not_null():
-    logger.info("Searching for NOT NULL constraint violations..."),
+    rslogger.info("Searching for NOT NULL constraint violations..."),
     not_null_count = 0
     async with async_session() as session:
         for table_name, table in Base.metadata.tables.items():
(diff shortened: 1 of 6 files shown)


bookserver/db.py
```diff
@@ -24,6 +24,8 @@
 from .config import settings, BookServerConfig, DatabaseType
 
 
+logger = logging.getLogger(__name__)
```
bnmnetp, line 27: I have a logger for us to use configured within the fastapi framework.  the routers all use this: `from ..applogger import rslogger`. Why not use the same here?

bjones1: Nice! I didn't realize that. I'll update -- thanks.

bjones1: I didn't use `rslogger` in the actual test code, since that seems appropriate, but it's now used in all bookserver code.

pyproject.toml
```diff
@@ -86,3 +86,5 @@ addopts = "--ignore=_build"
 env = [
     "BOOK_SERVER_CONFIG=test",
 ]
+# The _`default logging level` of INFO produces a LOT of output. Use a higher level of ``WARNING`` to reduce the noise.
+log_cli_level = 30
```
bnmnetp, line 90: is there no way to use `logging.WARNING`. It would be good to document what level logging is and how to override this from the cli.  It seems like a common workflow would be to run the tests, upon failure rerun the tests with the logging level set to debug to figure out what went wrong.

bnmnetp: I see that you did that below. sorry.

test/toctree.rst
```diff
@@ -1,7 +1,18 @@
 **************************
 Testing for the BookServer
 **************************
-To run the tests, execute ``poetry run pytest`` from the parent of this subdirectory; running them here causes code coverage failures.
+To run the tests, execute ``poetry run pytest`` from the parent of this subdirectory; running them here causes code coverage failures. There are several important command-line options:
+
+--skipdbinit    Skip initialization of the test database. This makes the tests start much faster, at the risk of a corrupt database causing spurious test failures.
+--server_debug  Enable server debug mode. This runs the server in a separate terminal/console, which allow you to set breakpoints, stop the code, etc.
+--log_cli_level LEVEL   Set the `pytest logging level <https://docs.pytest.org/en/6.2.x/logging.html#live-logs>`_. This level affects not just pytest, but the server and all tools run by the tests. Use ``--log_cli_level=INFO`` to provide complete output from the server; the `default logging level <default logging level>` is ``WARNING``.
```
bnmnetp, line 8: I think DEBUG is a lower level than INFO

bjones1: True, but then asyncpg (as I remember) spews incredible quantities of messages. There's probably a way to add a filter to only show messages from the bookserver, but I haven't look at that.

pyproject.toml
bjones1: I'd prefer to use `logging.WARNING`, bur got an error from pytest when I did. Sigh, that would be better.

bnmnetp (approved):

PR #67, opened 2021-08-12
Fix: Run CI on Postgres.

bjones1:
Aack! I thought we were already running CI using Postgres....but that wasn't true. It is now.


bnmnetp:
This seems to duplicate all of your logging work in a previous PR??

bjones1:
Yes, I left it a draft for this reason -- fixes here needed fixes from #66 as well. I'd suggest merging that first, then I'll update this.

bjones1:
**Important note**: I had to [remove the index from the courses table](https://github.com/bnmnetp/BookServer/pull/67/files#diff-5fc395447ecbf46aa65df0cc2881c0e603a221f0f7c494216654af3ddefd6822L348) to avoid weird Postgres error. I suggest we merge this first then investigate what's going on. Other than that, the changes are mainly to the test setup.

bjones1:
@bnmnetp, this is ready for your review.

bnmnetp:
Interesting, I'll bet that postgres creates an index to enforce the uniqueness constraint.

bnmnetp (approved):

PR #68, opened 2021-08-12
Add: Run using gunicorn if possible.

bjones1:


pyproject.toml
```diff
@@ -24,6 +24,8 @@ python = "^3.7"
 fastapi = "^0.68.0"
 # Per the `uvicorn docs <https://www.uvicorn.org/#quickstart>`_, install the standard (as opposed to minimal) uvicorn dependencies.
 uvicorn = {extras = ["standard"], version = "^0.14.0"}
+# See the `poetry docs <https://python-poetry.org/docs/dependency-specification/#using-environment-markers>`_.
+gunicorn = {version = "^20.1.0", markers = "sys.platform != 'win32'"}
```
bnmnetp, line 28: I'm confused.  My understanding is that uvicorn is the async version of gunicorn.  How does this actually work?  Has gunicorn become async?  Should we stick with uvicorn or always use gunicorn?

bnmnetp: I guess gunicorn can now use uvicorn as a worker.  That seems OK for windows dev and testing but seems less than ideal for production.  So I think we stick with uvicorn.

bnmnetp (approved):

pyproject.toml
bjones1: Uvicorn is a multi-threaded, but single-process, ASGI server. Gunicorn is a multi-process server. Nginx provides fast serving of static files, fast HTTPS connections, load balancing, etc. So, to quote the docs from [uvicorn deployment](https://www.uvicorn.org/deployment/):

> Server deployment is a complex area, that will depend on what kind of service you're deploying Uvicorn onto.
>
> As a general rule, you probably want to:
>
> - Run uvicorn --reload from the command line for local development.
> - Run gunicorn -k uvicorn.workers.UvicornWorker for production.
> - Additionally run behind Nginx for self-hosted deployments.
> - Finally, run everything behind a CDN for caching support, and serious DDOS protection.

PR #69, opened 2021-08-21
Logout

bjones1:
Add a logout endpoint, enable more tests, misc cleanup.

### DIFF
--- a/bookserver/routers/auth.py
+++ b/bookserver/routers/auth.py
@@ -109,6 +109,15 @@ async def login(
     return response
 
 
+# To log out, simply delete the cookie containing auth information.
+@router.get("/logout")
+async def logout():
+    # Send the user to the login page after the logout.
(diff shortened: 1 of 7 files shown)


bjones1:
@bnmnetp, this is ready for review.

bookserver/routers/auth.py
```diff
@@ -109,6 +109,15 @@ async def login(
     return response
 
 
+# To log out, simply delete the cookie containing auth information.
+@router.get("/logout")
```
bnmnetp, line 113: Nit: maybe add response_class: RedirectResponse

bnmnetp (approved):

bookserver/routers/auth.py
bjones1: Thanks! I'll add this.

PR #70, opened 2021-08-23
WIP: deploy using nginx

bjones1:
I won't be able to look at this until tomorrow at the earliest, since since is my big teaching day (plus tomorrow morning). Here's my (small) amount of progress, in case you want to play, along with my approach / plans.

- [ ] Handle HTTPS connections; forward HTTP to web2py/fastapi
- [ ] Serve static files -- the existing regexes for web2py should be fine. There's no regex possible for fastAPI, since the URLs refer to the course name instead of the base course name.
- [ ] Mount gnunicorn, which runs uvicorn, which runs FastAPI, which runs the BookServer at `/fa`. We already have code to run gunicorn + uvicorn + fastAPI + BookServer (in `__main__.py`). The top half of the config file was copied from gunicorn deployment docs.
- [ ] Mount uwsgi and its web2py connection at the `/` (root) directory. Hopefully, this means just using the existing nginx config (at the bottom of the `nginx.conf` file in this PR.

### DIFF
--- a/bookserver/nginx.conf
+++ b/bookserver/nginx.conf
@@ -1,0 +1,109 @@
+# ***********************************
+# |docname| - configuration for nginx
+# ***********************************
+# This sets up nginx to run both the old and new server together.
+#
+# It was partially taken from the `gunicorn docs <https://docs.gunicorn.org/en/stable/deploy.html#nginx-configuration>`_.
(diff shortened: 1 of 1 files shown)


bjones1:
OK, a first pass of this is ready and (sort of) working).

PR #71, opened 2021-08-27
Fix: add missing python3 field.

bjones1:
This fixes https://github.com/RunestoneInteractive/RunestoneServer/pull/1746 on the bookerver side.


bjones1:
Sounds good. I think that's a solid approach. I did some poking, and it doesn't seem related to the Web2PyBoolean, but instead to the default value.

bjones1:
Hopefully this will work...

bnmnetp (approved):

PR #73, opened 2021-08-31
Fix: Missing activity counts for page

bnmnetp:
I discovered that we had not implemented the activity counts for book pages so they always restart at 0.  This PR fixes that.

### DIFF
--- a/bookserver/crud.py
+++ b/bookserver/crud.py
@@ -15,13 +15,13 @@
 import datetime
 import json
 from collections import namedtuple
-from typing import List, Optional
+from typing import List, Optional, Dict
 
 # Third-party imports
 # -------------------
 from fastapi.exceptions import HTTPException
 from pydal.validators import CRYPT
-from sqlalchemy import and_, func, update
+from sqlalchemy import and_, func, update, distinct
 from sqlalchemy.sql import select, text
 
 from . import schemas
@@ -116,6 +116,39 @@ async def count_useinfo_for(
         res = await session.execute(query)
         rslogger.debug(f"res = {res}")
         return res.all()
+
+
+async def fetch_page_activity_counts(
+    chapter: str, subchapter: str, base_course: str, course_name: str, username: str
+) -> Dict[str, int]:
+    query = select(Question).where(
+        (Question.subchapter == subchapter)
+        & (Question.chapter == chapter)
+        & (Question.from_source == True)  # noqa: E712
+        & ((Question.optional == False) | (Question.optional == None))
+        & (Question.base_course == base_course)
+    )
+    async with async_session() as session:
+        page_divids = await session.execute(query)
+    rslogger.debug(f"PDVD {page_divids}")
+    div_counts = {q.name: 0 for q in page_divids.scalars()}
+    query = select(distinct(Useinfo.div_id)).where(
+        (Question.subchapter == subchapter)
+        & (Question.chapter == chapter)
+        & (Question.base_course == base_course)
+        & (Question.from_source == True)  # noqa: E712
+        & ((Question.optional == False) | (Question.optional == None))
+        & (Question.name == Useinfo.div_id)
+        & (Useinfo.course_id == course_name)
+        & (Useinfo.sid == username)
+    )
+    async with async_session() as session:
+        sid_counts = await session.execute(query)
+
+    for row in sid_counts.scalars():
+        div_counts[row.div_id] = 1
+
+    return div_counts
 
 
 async def fetch_poll_summary(div_id: str, course_name: str) -> List[tuple]:
@@ -404,6 +437,7 @@ async def create_initial_courses_users():
                 downloads_enabled=False,
                 courselevel="",
                 institution="",
+                new_server=False,
             )
             await create_course(new_course)
         # make a user
--- a/bookserver/crud.py
+++ b/bookserver/crud.py
@@ -121,6 +121,12 @@ async def count_useinfo_for(
 async def fetch_page_activity_counts(
     chapter: str, subchapter: str, base_course: str, course_name: str, username: str
 ) -> Dict[str, int]:
+    """
+    Used for the progress bar at the bottom of each page.  This function
+    finds all of the components for a particular page (chaper/subchapter)
+    and then finds out which of those elements the student has interacted
+    with.  It returns a dictionary of {divid: 0/1}
+    """
     query = select(Question).where(
         (Question.subchapter == subchapter)
         & (Question.chapter == chapter)
--- a/bookserver/crud.py
+++ b/bookserver/crud.py
@@ -15,13 +15,13 @@
 import datetime
 import json
 from collections import namedtuple
-from typing import List, Optional, Dict
+from typing import Dict, List, Optional
 
 # Third-party imports
 # -------------------
 from fastapi.exceptions import HTTPException
 from pydal.validators import CRYPT
-from sqlalchemy import and_, func, update, distinct
+from sqlalchemy import and_, distinct, func, update
 from sqlalchemy.sql import select, text
 
 from . import schemas
@@ -127,23 +127,23 @@ async def fetch_page_activity_counts(
     and then finds out which of those elements the student has interacted
     with.  It returns a dictionary of {divid: 0/1}
     """
-    query = select(Question).where(
+
+    where_clause_common = (
         (Question.subchapter == subchapter)
         & (Question.chapter == chapter)
         & (Question.from_source == True)  # noqa: E712
         & ((Question.optional == False) | (Question.optional == None))
         & (Question.base_course == base_course)
     )
+
+    query = select(Question).where(where_clause_common)
+
     async with async_session() as session:
         page_divids = await session.execute(query)
     rslogger.debug(f"PDVD {page_divids}")
     div_counts = {q.name: 0 for q in page_divids.scalars()}
     query = select(distinct(Useinfo.div_id)).where(
-        (Question.subchapter == subchapter)
-        & (Question.chapter == chapter)
-        & (Question.base_course == base_course)
-        & (Question.from_source == True)  # noqa: E712
-        & ((Question.optional == False) | (Question.optional == None))
+        where_clause_common
         & (Question.name == Useinfo.div_id)
         & (Useinfo.course_id == course_name)
         & (Useinfo.sid == username)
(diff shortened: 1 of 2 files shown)


bookserver/crud.py
```diff
@@ -118,6 +118,39 @@ async def count_useinfo_for(
...
 
 
+async def fetch_page_activity_counts(
+    chapter: str, subchapter: str, base_course: str, course_name: str, username: str
+) -> Dict[str, int]:
+    query = select(Question).where(
+        (Question.subchapter == subchapter)
+        & (Question.chapter == chapter)
+        & (Question.from_source == True)  # noqa: E712
+        & ((Question.optional == False) | (Question.optional == None))
```
bjones1, line 128: Nitpick: isn't this `Question.optional != True`?

bookserver/crud.py
```diff
@@ -118,6 +118,39 @@ async def count_useinfo_for(
         return res.all()
 
 
+async def fetch_page_activity_counts(
+    chapter: str, subchapter: str, base_course: str, course_name: str, username: str
+) -> Dict[str, int]:
+    query = select(Question).where(
```
bjones1, line 124: Add a comment, perhaps "Find all questions on the given page (defined as a specific chapter and subchapter of a book)".

bookserver/crud.py
```diff
@@ -118,6 +118,39 @@ async def count_useinfo_for(
...
+        & (Question.chapter == chapter)
+        & (Question.from_source == True)  # noqa: E712
+        & ((Question.optional == False) | (Question.optional == None))
+        & (Question.base_course == base_course)
+    )
+    async with async_session() as session:
+        page_divids = await session.execute(query)
+    rslogger.debug(f"PDVD {page_divids}")
+    div_counts = {q.name: 0 for q in page_divids.scalars()}
+    query = select(distinct(Useinfo.div_id)).where(
```
bjones1, line 135: Add comment, perhaps "Find all activity on the specified page given a user and course." Also, it seems that the query below could be DRYer if defined as `query1 = query & (Question.name == Useinfo.div_id) & ...`.

bookserver/crud.py
```diff
@@ -404,6 +437,7 @@ async def create_initial_courses_users():
                 downloads_enabled=False,
                 courselevel="",
                 institution="",
+                new_server=False,
```
bjones1, line 439: Why this? I don't understand. I thought polls could run on either the bookserver or the web2py server.

bookserver/routers/books.py
```diff
@@ -110,19 +111,28 @@ async def serve_page(
...
         logged_in = "false"
-
+        activity_info = {}
     course_row = await fetch_course(user.course_name)
     if not course_row:
         raise HTTPException(status_code=404, detail=f"Course {course_name} not found")
     rslogger.debug(f"Base course = {course_row.base_course}")
+    chapter = os.path.split(os.path.split(pagepath)[0])[1]
+    subchapter = os.path.basename(os.path.splitext(pagepath)[0])
+    if logged_in:
```
bjones1, line 124: This is always true. Use `if user:` instead. I'd also suggest putting it in the `if user:` clause above.

bnmnetp:
Strange test results...  Everything except  the lp tests pass locally for me.  Am I missing a dependency?  It just times out every time I run the test, but all others pass consistently.

bookserver/routers/books.py
```diff
@@ -152,13 +162,13 @@ async def serve_page(
...
-        user_email="bonelake@mac.com",
+        user_email=user.email if logged_in else "",
         downloads_enabled="false",
         allow_pairs="false",
-        activity_info={},
+        activity_info=json.dumps(activity_info),
         settings=settings,
         is_logged_in=logged_in,
-        is_instructor="true",
+        is_instructor="true" if user_is_instructor else "false",
```
bjones1, line 171: Nitpick: this is `json.dumps(user_is_instructor)`.

bookserver/routers/books.py
```diff
@@ -110,19 +111,28 @@ async def serve_page(
     # need to use lowercase true and false.
     if user:
         logged_in = "true"
+        user_is_instructor = await is_instructor(request)
```
bjones1, line 114: Define this as `False` in the `else` branch.

bookserver/routers/books.py
```diff
@@ -110,19 +111,28 @@ async def serve_page(
     # need to use lowercase true and false.
     if user:
         logged_in = "true"
```
bjones1, line 113: Nitpick: use `json.dumps(bool(user))`.

bjones1 (changes requested):
I saw a few logic errors that need fixing; otherwise, good progreess!

bjones1:
> Strange test results... Everything except the lp tests pass locally for me. Am I missing a dependency? It just times out every time I run the test, but all others pass consistently. Same behavior on main as on this PR branch, so I'm pretty confident this PR is fine.

I take a look. My guess is that redis isn't running.

bookserver/crud.py
bnmnetp: I'm never sure how it will interpret null. this may be a hold over from weird DAL behavior.

bookserver/crud.py
bnmnetp: The new_server variable is only important in the context of web2py where we would be routing the book requests for this course to /ns

bookserver/routers/books.py
bnmnetp: tricky, but I think less obvious.

bookserver/routers/books.py
bnmnetp: oops

bnmnetp: we need course information first, so I put it here

bookserver/crud.py
bjones1: Probably, since in `models.py` the field is allowed to be null.

bookserver/crud.py
bjones1: Still lost, but if it makes sense to you, I'm happy.

bookserver/routers/books.py
bjones1: Sounds good.

bookserver/routers/books.py
bjones1: Yes, I see that now...

bookserver/crud.py
bnmnetp: I don't think that works given the difference in the select clause.

bookserver/crud.py
bnmnetp: This was my solution so that we could keep most courses running on the web2py server, but have a few beta testers try running their courses on the new server.  

What do polls have to do with it??

bookserver/crud.py
bjones1: True. My revised (and untested) suggestion:

``` Python
q_where = (
    (Question.subchapter == subchapter)
    & (Question.chapter == chapter)
    & (Question.from_source == True)  # noqa: E712
    & ((Question.optional != True)
    & (Question.base_course == base_course)
)
query = select(Question).where(q_where)
...
query = select(distinct(Useinfo.div_id)).where(
    q_where
    & (Question.name == Useinfo.div_id)
    & (Useinfo.course_id == course_name)
    & (Useinfo.sid == username)
)
```

I like this because it clearly shows that the second query is the same as the first but with extra criteria.

bookserver/crud.py
bjones1: I don't know, which is what confuses me. But this is in the `fetch_poll_summary` function, which is called by the `/getpollresults` endpoint, so I assume polls matter for this in some way...

bookserver/routers/books.py
```diff
@@ -152,13 +164,13 @@ async def serve_page(
         # _`root_path`: The server is mounted in a different location depending on how it's run (directly from gunicorn/uvicorn or under the ``/ns`` prefix using nginx). Tell the JS what prefix to use for Ajax requests. See also `setting root_path <setting root_path>` and the `FastAPI docs <https://fastapi.tiangolo.com/advanced/behind-a-proxy/>`_. This is then used in the ``eBookConfig`` of :doc:`runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html`.
         new_server_prefix=request.scope.get("root_path"),
         # TODO
-        user_email="bonelake@mac.com",
+        user_email=user.email if logged_in else "",
```
bjones1, line 167: Should be `if user` instead, since `logged_in` is always true.

bjones1 (approved):
I noticed just one more logic error -- make sure to fix that. The rest is choice/preference, so I'll leave that up to you.

bjones1:
Re: test failures, the only failure I see is in test_toc_decorators, not test_lp_1. ???

I also see a lot of strange JavaScript confusion in the error logs. What's going on with Runestone Components?

bookserver/crud.py
bnmnetp: That is definitely a different style.... I followed your suggestion on this one.

bookserver/crud.py
bjones1: OK, my apologies. I look at the code in github completely wrong. Sigh.

bookserver/crud.py
bjones1: Great!

bnmnetp: I'm very skeptical of the != True part given how it is represented as a character in Postgres.  Maybe I'll play around with it in my spare time to see if it behaves.

bnmnetp: A quick query from the command line

`select * from courses where allow_pairs != 'T'` ONLY returns rows where it is 'F' It does not return rows that have NULL in the allow_pairs field.

bjones1: Sounds good. The web2pybool class translates the characters / null to True, False, or None. If it doesn't, something's wrong.

bookserver/routers/books.py
bnmnetp: got it! Thanks.

bnmnetp:
Yes, that is what I'm saying, when I run the tests locally they all pass except for LP, but on GitHub we are seeing some oddness with javascript and test_toc_decorators. I suspect some configuration oddness.

bnmnetp:
OK, user error on my part, I forgot to push the changes to components bookserver branch that go with this.

Good catch.

Too many branches.....  Can't wait until we fully adopt bookserver.

PR #75, opened 2021-09-10
Add: new indices

bnmnetp:
I recently spent a bunch of time tracking down a major performance issue in the autograder.  The solution was these new indices.  From 10 seconds + per student down to 25 seconds total for a class of 161!


bjones1:
Wow, this is really exciting! Do you have any notes that explain how you tracked this down, in case the autograder gets slow again?

PR #77, opened 2021-09-21
Discussions -- peer instruction

bnmnetp:
OK, 

I think I've lived with this long enough... and it would be great to remove at least one branch.

I would like to merge all three branches (RunestoneServer, Components, and here) and assign a few courses to use new server.

I've been using this exclusively on dev and I'm comfortable with putting this new server into production.


bjones1:
Sounds great! I'm definitely in favor or merging branches and reducing all this confusion.

Do you want me to review this PR, then? Do you want any help getting the tests to pass?

bnmnetp:
I'll do a little more cleanup of comments but it should be ready for review.


All tests are passing locally.  The problem seems to be with an update to FastAPI-Login update from 1.6.3 to 1.7.0 breaks everything and causes weird mypy failures.

For the moment I'm pinning FastAPI-Login to 1.6.3 but that will need some investigation to figure out what is going on.

bookserver/routers/discuss.py
```diff
@@ -1,14 +1,63 @@
+# ******************************************************
```
bjones1, line 1: I love these docs!!! Very nice.

IMHO, don't manually wrap paragraphs -- that makes editing hard later. I think https://queue.acm.org/detail.cfm?id=1053354 makes a good case.

bookserver/routers/discuss.py
```diff
@@ -21,19 +70,162 @@
...
-    while True:
-        data = await websocket.receive_text()
-        await websocket.send_text(f"Message sent was {data}")
+class ConnectionManager:
+    def __init__(self):
+        self.active_connections: Dict[str, WebSocket] = {}
+
+    async def connect(self, user: str, websocket: WebSocket):
+        await websocket.accept()
+        self.active_connections[user] = websocket
```
bjones1, line 79: Perhaps `assert user not in active_connections`?

bookserver/routers/discuss.py
```diff
@@ -21,19 +70,162 @@
...
+    The browser and the server.  The websocket is persistent as long
+    as the process that runs this endpoint is alive.
+
+    Note: This function must return which means it must be throroughly async
+    Using a non async library like plain redis-py will not work as the subscriber
+    will block
+    """
+    rslogger.debug(f"IN WEBSOCKET {uname=}")
+    username = uname
+    local_users.add(username)
```
bjones1, line 148: You might add `global local_users` just to make it clear where this var comes from. (Not required, but helps explain.)

bookserver/routers/discuss.py
```diff
@@ -21,19 +70,162 @@
...
+    await manager.connect(username, websocket)
+    r = aioredis.from_url(os.environ.get("REDIS_URI", "redis://localhost:6379/0"))
```
bjones1, line 150: I'd consider putting config values like this in `config.settings`.

bookserver/routers/discuss.py
```diff
@@ -21,19 +70,162 @@
...
+                else:
+                    rslogger.error("unknown message type {pmess['type']}")
+                    continue
+                if data["broadcast"]:
+                    await manager.broadcast(data)
+                else:
+                    # because **every** connection runs this same loop
+                    # we only want to send a non-broadcast message if
+                    # it is to ourself.
+                    partner = await r.hget("partnerdb", data["from"])
```
bjones1, line 201: Future thoughts: support more than 1 partner (get a list, instead of one item here).

bookserver/routers/discuss.py
```diff
@@ -21,19 +70,162 @@
...
+                    partner = partner.decode("utf8")
+                    if partner == username:
+                        await manager.send_personal_message(partner, data)
+                        # log the message
+                        await create_useinfo_entry(
+                            UseinfoValidation(
+                                event="sendmessage",
+                                act=f"to:{partner}:{data['message']}",
```
bjones1, line 209: A long message will cause a validation exception here -- perhaps send only the first 500 chars of the message?

bookserver/routers/discuss.py
```diff
@@ -21,19 +70,162 @@
...
+                )
+                # TODO: now that we have multi-await working it *may* be more
+                # efficient to go back to sending all peer messages by websocket
+                # and putting them into the subscriber queue here rather than
+                # having a separate endpoint.
+
+
+@router.post("/send_message")
+async def send_message(packet: PeerMessage):
+    r = await aioredis.from_url(os.environ.get("REDIS_URI", "redis://localhost:6379/0"))
```
bjones1, line 230: DRY this, use `config.settings`.

bjones1 (approved):

bookserver/routers/discuss.py
bnmnetp: but super hard to read in the editor.   I don't like having word wrap on for code.

bookserver/routers/discuss.py
bnmnetp: I think that could potentially make a race condition.  When a user refreshes the page the web socket closes and then reopens with a call (potentially) to the same server.    The server will also note this and remove the user from the active connections.  I think the assert just needlessly creates a new fail condition.

bookserver/routers/discuss.py
bnmnetp: safety first!

bookserver/routers/discuss.py
bnmnetp: Yep, that came up in the call the other day, but can definitely be a future enhancement.

bookserver/routers/discuss.py
bjones1: OK, sounds good.

bookserver/routers/discuss.py
bjones1: I like black in this case -- it word wraps code for me, then my editor word wraps the text. To each his own :).

PR #78, opened 2021-10-01
Peer tweaks

bjones1:
Fixes per our discussion on Slack.


bjones1:
Tests fail due to mypy complaining about something in the peer discussion code -- @bnmnetp, I'll let you fix that.

bnmnetp (approved):

PR #91, opened 2022-01-14
Fix: Correct LP answer column type.

bjones1:
This now agrees with the type from web2py / what's in the database.

### DIFF
--- a/bookserver/models.py
+++ b/bookserver/models.py
@@ -290,9 +290,9 @@ UnittestAnswersValidation = sqlalchemy_to_pydantic(UnittestAnswers)
 @register_answer_table
 class LpAnswers(Base, AnswerMixin):
     __tablename__ = "lp_answers"
-    # See answer_. A JSON string; see RunestoneComponents for details. TODO: The length seems too short to me. Migrate this to use a ``Text`` field type instead.
-    answer = Column(String(512), nullable=False)
-    # A grade between 0 and 100. None means it the student hasn't submitted an answer yet. This was added before the ``percent`` field most other question types now have; it servers the same role, but stores the answer as a percentage. (The ``percent`` field in other questions stores values between 0 and 1.)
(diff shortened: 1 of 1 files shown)


bnmnetp (approved):

bookserver/models.py
```diff
@@ -290,9 +290,9 @@ class UnittestAnswers(Base, CorrectAnswerMixin):
 @register_answer_table
 class LpAnswers(Base, AnswerMixin):
     __tablename__ = "lp_answers"
-    # See answer_. A JSON string; see RunestoneComponents for details. TODO: The length seems too short to me. Migrate this to use a ``Text`` field type instead.
-    answer = Column(String(512), nullable=False)
-    # A grade between 0 and 100. None means it the student hasn't submitted an answer yet. This was added before the ``percent`` field most other question types now have; it servers the same role, but stores the answer as a percentage. (The ``percent`` field in other questions stores values between 0 and 1.)
+    # See answer_. A JSON string; see RunestoneComponents for details.
```
bnmnetp, line 293: Hmmm, should PR's such as this on have an accompanying alembic migration?

bnmnetp (approved):

bookserver/models.py
bjones1: Oddly, my database had the correct `text` type already, even though I'd used Alembic to migrate from web2py to bookserver's models. Perhaps alembic can't distinguish between a `varchar` and `text`? I ran Alembic, and it detected no changes as well. What type does do you see for answers column in `psql -c "\d lp_answers" $DBURL`?

bnmnetp: My database shows `text` as the type for answer.

bnmnetp: Oddly, useinfo act is declared as `varchar(512)`.   (NOT ARCHER autocorrect!!)

bjones1: :) It would make more sense to have `act` be `text`...

PR #93, opened 2022-02-02
Add: Update versions of dependencies.

bjones1:


bjones1:
This is a first step in providing more flexible version requirements. I ran into an issue: tests fail if we upgrade from FastAPI v0.68 to v0.69 (which isn't even the current version -- it's at v0.73). The root cause: FastAPI v0.69 provides a different `TestClient` which isn't compatible with sqlalchemy's asyncio approach. Switching to an async `TestClient` is my best guess at fixing this, but it's work I'm postponing for a future PR.

bookserver/session.py
```diff
@@ -45,6 +46,10 @@ async def load_user(user_id: str):
     return await fetch_user(user_id)
 
 
+# The ``user_loader`` decorator doesn't propagate type hints. Fix this manually.
+load_user = cast(Callable[[str], Awaitable[AuthUserValidator]], _load_user)
```
bnmnetp, line 50: Ugh!!  This is where mypy goes completely off the rails IMHO.  Something that should not get in your way but add value by ensuring type correctness has now made the code very obscure.

pyproject.toml
```diff
@@ -26,48 +26,48 @@ celery = {extras = ["redis"], version = "^5.1.2"}
...
 # Per the `uvicorn docs <https://www.uvicorn.org/#quickstart>`_, install the standard (as opposed to minimal) uvicorn dependencies.
-uvicorn = {extras = ["standard"], version = "^0.14.0"}
+uvicorn = {extras = ["standard"], version = "^0.17.0"}
 # See the `poetry docs <https://python-poetry.org/docs/dependency-specification/#using-environment-markers>`_.
-gunicorn = {version = "^20.1.0", markers = "sys.platform != 'win32'"}
-Jinja2 = "^3.0.1"
-aiofiles = "^0.6.0"
-alembic = "^1.4.3"
-python-dateutil = "^2.8.1"
-fastapi-login = "1.6.3"
```
bnmnetp, line 36: I'm not sure how much of fastapi-login we even use anymore....  But I'll need to test this carefully, login bumps have caused me too many headaches already.

pyproject.toml
```diff
@@ -26,48 +26,48 @@ celery = {extras = ["redis"], version = "^5.1.2"}
 python = "^3.8.0"
 fastapi = "^0.68.0"
 # Per the `uvicorn docs <https://www.uvicorn.org/#quickstart>`_, install the standard (as opposed to minimal) uvicorn dependencies.
-uvicorn = {extras = ["standard"], version = "^0.14.0"}
+uvicorn = {extras = ["standard"], version = "^0.17.0"}
 # See the `poetry docs <https://python-poetry.org/docs/dependency-specification/#using-environment-markers>`_.
-gunicorn = {version = "^20.1.0", markers = "sys.platform != 'win32'"}
-Jinja2 = "^3.0.1"
```
bnmnetp, line 32: It is good to get this syntax cleaned up, as it was not really doing what we thought it was doing.

bnmnetp (approved):

bookserver/session.py
bjones1: I agree, though part of the blame goes to FastAPI-LoginManager, since that's what fails to propagate type hints.

pyproject.toml
bjones1: Sounds good. I agree that manual testing is always important, especially since our test coverage is < 50%.

pyproject.toml
bjones1: Good point, thanks!

pyproject.toml
bjones1: I did some quick manual testing -- everything works for me. I'm using the web2py-based login, since we rarely log in using the BookServer.

PR #98, opened 2022-02-23
Fix: Provide server-side feedback on the /results 

bjones1:
(get_assessment_results) endpoint.

### DIFF
--- a/bookserver/routers/assessment.py
+++ b/bookserver/routers/assessment.py
@@ -15,22 +15,21 @@
 #
 # Standard library
 # ----------------
-import random
 import datetime
+import random
 from typing import Optional, Dict, Any
-
-# import pdb
 
 # Third-party imports
 # -------------------
+from bleach import clean
 from fastapi import APIRouter, HTTPException, Request, status
 from pydantic import BaseModel
-from bleach import clean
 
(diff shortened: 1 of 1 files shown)


bookserver/routers/assessment.py
```diff
@@ -70,38 +71,42 @@ async def get_assessment_results(
...
-    #   do_server_feedback, feedback = is_server_feedback(div_id, course)
-    #   if do_server_feedback:
-    #       correct, res_update = fitb_feedback(rows.answer, feedback)
-    #       res.update(res_update)
-    rslogger.debug(f"Returning {row}")
-    return make_json_response(detail=row)
+    ret = row.dict()
+
+    # Do server-side grading if needed, which restores the answer and feedback.
+    if feedback := await is_server_feedback(request_data.div_id, request_data.course):
```
bnmnetp, line 101: Could a flag be provided as part of the AssessmentRequest that indicates if server side grading is used for this question?  -- This is the 3rd most called endpoint so I want to make sure we consider every additional database hit.

bjones1: Let me think about this -- good point.

bjones1:
Great, thanks!

PR #102, opened 2022-06-30
Celery move

bjones1:
After merge, I'll make a matching PR on the Runestone Server repo that deletes these files there.


bjones1:
This is ready -- the Literate Programming test will fail until the server-side PR gets merged. I want to merge this PR first (add files) before the server-side PR (delete files), though, since that makes more sense to me.

PR #111, opened 2022-10-20
Adhoc practice

bnmnetp:
This PR restores functionality lost in the initial bookserver port.  Namely the option for the practice feature to automatically add questions for a student as they complete each page.

* This moves schema ownership for course_practice and user_topic_prace tables to the BookServer
* Next step is to port the practice page app to the bookserver


bnmnetp:
Merged from CL

PR #112, opened 2022-11-06
Fix: Update Alembic with database tables now owned by the bookserver.

bjones1:
See also https://github.com/RunestoneInteractive/RunestoneServer/pull/2025.


bjones1:
Tests fail because they need fixing, not because of this PR.

bnmnetp (approved):

PR #113, opened 2022-11-08
Fix: Merge the two current heads in Alembic.

bjones1:


alembic/versions/217f49a3b6be_merge_heads.py
```diff
@@ -0,0 +1,27 @@
+"""Merge heads
```
bnmnetp, line 1: Are you merging this by hand?  As it stands this does nothing so there is no reason to even keep it?

bjones1: No, this is generated from the CLI using `alembic merge heads`. Currently, we have two branches in Alembic. This just merges the two branches together from Alembic's perspective, but doesn't actually change the DB. It's clean-up of some Alembic mess (multiple heads), not deep changes. We need to keep it (and you need to run `alembic update head`) so that future migrations get built on this merge, instead of creating more branches.

bjones1: After you run this, I'm curious to see what a `alembic revision --autogenerate -m "Update to current db structure"` produces for you. It does create some operations for me, but perhaps my db is missing something?

bjones1:
Yes, I'll send more PRs after this is merged.

bnmnetp (approved):

bjones1:
Given the failing tests, you'll have to merge this -- I don't have access to push.

PR #115, opened 2022-11-10
Fix: Correct type declaration.

bjones1:


bookserver/crud.py
```diff
@@ -1123,7 +1123,7 @@ async def create_user_topic_practice(
 
 async def fetch_qualified_questions(
     base_course, chapter_label, sub_chapter_label
-) -> list[QuestionValidator]:
```
bnmnetp, line 1126: This is not incorrect it just requires python > 3.8.  I think it is asking the question what do we want to support?

3.7.0 - June 2018
3.8.0 - Oct 2019
3.9.0 - Oct 2020
3.10.0 - Oct 2021
3.11.0 - Oct 2022

see https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc

bjones1: Good point, I didn't realize that.

Do we want to pick some sort of metric for what Python versions to support? For example, the last three releases?

Supporting until a version of Python reaches its [end of life](https://endoflife.date/python) means Python 3.7 isn't yet EOL (ack!) and Python 3.8 lasts almost 2 more years...not really what I want to do...

bnmnetp: I think it is reasonable for us to specify a minimum version that we want to support.  I think 3.9 is pretty reasonable, especially given that we specify the version in the Dockerfile.

bjones1: Good point. I'm closing this in favor of #116, where we run tests on Python 3.9 and 3.10 instead of 3.8 and 3.9.

PR #117, opened 2022-12-06
Add: Copy over a Verilog test bench if present.

bjones1:


bjones1:
Thanks! Would you merge this? Since test fail, I don't have permission to merge.

PR #119, opened 2023-03-21
Python check endpoint

ascholerChemeketa:
Server endpoint to get pyflakes output from code sample

Question:
Are you happy with the endpoint being in /ns/books/?
Should it be in some new group? (/ns/utils/python_check/ or something like that)

### DIFF
--- a/bookserver/main.py
+++ b/bookserver/main.py
@@ -36,6 +36,7 @@ from .routers import auth
 from .routers import books
 from .routers import rslogging
 from .routers import discuss
+from .routers import coach
 from .session import auth_manager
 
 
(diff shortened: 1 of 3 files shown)


bnmnetp:
when you asked before I was not thinking about this particular thing as having a home in books.  
I am thinking that we might want to put this in its own file under routers.   Maybe `coach.py`
`/ns/coach/python_check` ?

ascholerChemeketa:
Yes, that makes sense. Will move and update PR.

ascholerChemeketa:
Endpoint moved

bnmnetp (approved):

bnmnetp:
@ascholerChemeketa Can you run `black` on coach.py please?

bnmnetp:
Nevermind, I can do it just as easy.

